PR #1, opened 2026-04-09
Fix: ld: warning & Make make build as default target & Print effective config in startup & Add AGENTS.md

xxx7xxxx:


third_party/boxlite-go/bridge_cgo_dev.go
```diff
@@ -10,7 +10,7 @@ package boxlite
 
 #cgo darwin LDFLAGS: ${SRCDIR}/../../target/debug/libboxlite.a
 #cgo darwin LDFLAGS: -framework CoreFoundation -framework Security -framework IOKit
-#cgo darwin LDFLAGS: -framework Hypervisor -framework vmnet -lresolv
+#cgo darwin LDFLAGS: -framework Hypervisor -framework vmnet
```
RussellLuo, line 13: Any issues with this option?

cli/serve.go
```diff
@@ -85,6 +85,7 @@ func (a *App) runInternalServe(ctx context.Context, args []string, globals Globa
 		cfg.Server.APIBaseURL = strings.TrimRight(globals.Endpoint, "/")
 	}
 
+	a.printEffectiveConfig(cfg)
```
RussellLuo, line 95: All settings will be written in `~/.csgclaw/config.toml` (after `csgclaw onboard`).

xxx7xxxx: There could be multiple sources such as envs, flags, config file and so on. So I added the display to show the final effective config.

third_party/boxlite-go/bridge_cgo_dev.go
xxx7xxxx: No bugs, but it's deprecated and useless on Mac, My local go1.26.1 outputed warning. (BTW we should update the go version to the latest one go1.26)

RussellLuo: > update the go version to the latest one go1.2

LGTM, go ahead.

cli/serve.go
RussellLuo: Got it. Then should we mask the api_key with asterisks?

third_party/boxlite-go/bridge_cgo_dev.go
xxx7xxxx: done

cli/serve.go
xxx7xxxx: done

PR #3, opened 2026-04-17
Normalize PicoClaw bridge model IDs

xxx7xxxx:
## Normalize PicoClaw bridge model IDs

- Prefix PicoClaw provider-facing model IDs with `openai/` when routing through the local OpenAI-compatible bridge.
- Preserve raw model names for CSGClaw profile/model display and default agent selection.
- Update the embedded PicoClaw config template and env var tests to cover slash-containing model IDs like `Qwen/Qwen3-0.6B-GGUF`.

### DIFF
--- a/internal/agent/defaults/picoclaw-config.json
+++ b/internal/agent/defaults/picoclaw-config.json
@@ -9,7 +9,6 @@
       "restrict_to_workspace": false,
       "allow_read_outside_workspace": true,
       "provider": "",
-      "model": "minimax-m2.7",
       "model_name": "minimax-m2.7",
       "max_tokens": 32768,
       "max_tool_iterations": 50,
--- a/internal/agent/defaults/picoclaw-config.json
+++ b/internal/agent/defaults/picoclaw-config.json
@@ -9,6 +9,7 @@
       "restrict_to_workspace": false,
       "allow_read_outside_workspace": true,
       "provider": "",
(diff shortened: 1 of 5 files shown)


internal/agent/defaults/picoclaw-config.json
```diff
@@ -9,6 +9,7 @@
       "restrict_to_workspace": false,
       "allow_read_outside_workspace": true,
       "provider": "",
+      "model": "minimax-m2.7",
```
RussellLuo: There is no "model" field in `agents.defaults` (see [AgentDefaults](https://github.com/sipeed/picoclaw/blob/0da962c4b46f75b392728c01422eab1202c85675/pkg/config/config.go#L254C6-L277))?

xxx7xxxx: fixed

internal/agent/defaults/picoclaw-config.json
```diff
@@ -73,7 +73,7 @@
   "model_list": [
     {
       "model_name": "minimax-m2.7",
-      "model": "minimax-m2.7",
+      "model": "openai/minimax-m2.7",
```
RussellLuo, line 76: Heads-up: changing this won't take effect right now. Gotta rebuild the image.

Everything else LGTM.

PR #5, opened 2026-04-17
Integrate with CSGHub-lite

xxx7xxxx:
## Implemented CSGHub-lite integration for csgclaw onboard and serve.

- Added --provider csghub-lite|custom while preserving existing non-interactive flags.
- Made interactive onboard provider-aware: fresh configs default to CSGHub-lite, existing configs default to keeping the current provider.
- Added CSGHub-lite /v1/models discovery, configurable base URL prompt, and local defaults: http://127.0.0.1:11435/v1 + API key local.
- Added startup/preflight checks before serve creates the manager when the default provider is local CSGHub-lite.
- Added stdin abstraction for testable interactive prompts.
- Updated README/README.zh and Makefile onboard example.
- Added tests for interactive onboard flows, CSGHub-lite model discovery, config persistence, and serve preflight.

### DIFF
--- a/cli/app.go
+++ b/cli/app.go
@@ -27,6 +27,7 @@ const (
 )
 
 type App struct {
+	stdin      io.Reader
 	stdout     io.Writer
 	stderr     io.Writer
 	httpClient HTTPClient
@@ -44,6 +45,7 @@ type GlobalOptions struct {
 
 func New() *App {
 	app := &App{
+		stdin:      os.Stdin,
 		stdout:     os.Stdout,
 		stderr:     os.Stderr,
 		httpClient: &http.Client{},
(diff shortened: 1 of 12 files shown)


cli/app.go
```diff
@@ -278,8 +280,13 @@ func (g GlobalOptions) commandOptions() command.GlobalOptions {
 }
 
 func (a *App) commandContext() *command.Context {
+	stdin := a.stdin
+	if stdin == nil {
+		stdin = os.Stdin
+	}
```
RussellLuo: Shouldn't `a.stdin` always be `os.Stdin`?

cli/csgclawcli/app.go
```diff
@@ -246,8 +248,13 @@ func (g GlobalOptions) commandOptions() command.GlobalOptions {
 }
 
 func (a *App) commandContext() *command.Context {
+	stdin := a.stdin
+	if stdin == nil {
+		stdin = os.Stdin
+	}
```
RussellLuo: Same question here.

cli/app.go
xxx7xxxx: Updated. `New()` already initializes `a.stdin` to `os.Stdin`, so I removed the redundant nil fallback and now pass `a.stdin` directly in both CLI app variants.

cli/csgclawcli/app.go
xxx7xxxx: Ditto

RussellLuo:
LGTM

PR #11, opened 2026-04-23
fix(network): use relative paths to resolve cloud deployment issues

jialudev:
- Removed leading slashes from `fetch` and `EventSource` URLs in `app.js` (e.g., changed `/api/v1/...` to `api/v1/...`).
- Dynamically injected a `<base>` tag in `index.html` to handle base paths correctly.
- These changes ensure that both API requests and static assets resolve reliably across different cloud deployment environments, especially when the application is not hosted at the root domain.

### DIFF
--- a/web/static/app.js
+++ b/web/static/app.js
@@ -406,7 +406,7 @@ function App() {
   const messageListRef = useRef(null);
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


RussellLuo:
LGTM

PR #17, opened 2026-04-28
fix(serve): enforce default manager_image and bump tag

wanghaojie124:
When starting serve, prefer the code default manager_image over the config value. Also bump the default picoclaw image tag to 2026.4.27.0 across docs and Makefile.

Made-with: Cursor


RussellLuo:
LGTM

PR #24, opened 2026-04-29
Introduce auth for codex/claude code & Fix agent not receiving messages after server restarted

xxx7xxxx:
### Summary

- Added CLIProxy auth bridge for Codex and Claude Code, including Codex auth import, macOS Claude Keychain probing, auth status/login APIs, CLI auth commands, and UI auth states.
- Fixed agent reply loss after `csgclaw serve` restart by reconnecting/recreating agent gateways, retrying transient BoxLite runtime-lock failures, and replaying missed PicoClaw messages.
- Updated docs and tests for auth flows, restart recovery, and PicoClaw delivery.

### DIFF
--- a/cli/model/model.go
+++ b/cli/model/model.go
@@ -1,4 +1,4 @@
-package auth
+package model
 
 import (
 	"bufio"
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -77,7 +77,7 @@ Top-level commands:
 - `serve`
 - `stop`
 - `agent`
-- `auth`
+- `model`
 - `user`
 - `bot`
 - `room`
@@ -166,14 +166,14 @@ Behavior:
 - Sends `SIGTERM` to the PID stored in the PID file.
 - If the process is already gone, it removes the stale PID file and reports that state.
 
-### `csgclaw auth`
+### `csgclaw model auth`
 
-Manages local Codex and Claude Code authentication for the embedded CLIProxyAPI.
+Manages local Codex and Claude Code authentication for model providers through the embedded CLIProxyAPI.
 
 Usage:
 
 ```bash
-csgclaw auth login <provider> [flags]
+csgclaw model auth login <provider> [flags]
 ```
 
 Providers:
@@ -190,13 +190,13 @@ Behavior:
 - `codex` first reuses `~/.codex/auth.json` when available, then starts OAuth if needed.
 - `claude-code` first probes macOS Keychain when available, then starts OAuth if needed.
 - Auth is stored in the CLIProxy auth directory, defaulting to `~/.cli-proxy-api`.
-- No top-level aliases such as `csgclaw codex-login` or `csgclaw claude-login` are registered.
+- Model provider auth is scoped under `csgclaw model auth`, not the server's own API authentication.
 
 Examples:
 
 ```bash
-csgclaw auth login codex
-csgclaw auth login claude-code --no-browser
+csgclaw model auth login codex
+csgclaw model auth login claude-code --no-browser
 ```
 
 ### `csgclaw agent`
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -77,6 +77,7 @@ Top-level commands:
 - `serve`
 - `stop`
 - `agent`
+- `auth`
 - `user`
 - `bot`
 - `room`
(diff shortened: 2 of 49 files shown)


internal/api/picoclaw.go
```diff
@@ -90,6 +97,79 @@ func (h *Handler) handlePicoClawEvents(w http.ResponseWriter, r *http.Request, b
 	}
 }
 
+func (h *Handler) replayRecentPicoClawMessages(botID string) {
```
RussellLuo: How can we ensure that replayed messages do not get duplicated before and after the server restarts?

docs/cli.md
```diff
@@ -165,6 +166,39 @@ Behavior:
...
 - If the process is already gone, it removes the stale PID file and reports that state.
 
+### `csgclaw auth`
+
+Manages local Codex and Claude Code authentication for the embedded CLIProxyAPI.
+
+Usage:
+
+```bash
+csgclaw auth login <provider> [flags]
```
RussellLuo: Since it has nothing to do with csgclaw's own auth, it is suggested to change the command to: `csgclaw model auth login`.

Reference: [OpenClaw Auth profiles](https://docs.openclaw.ai/cli/models#auth-profiles).

internal/api/picoclaw.go
xxx7xxxx: Fixed: Replay finds that message ID in persisted IM history, uses its timestamp as a cursor, and skips messages at or before it.

docs/cli.md
xxx7xxxx: My mind: too ugly long. 
My hand: fix without question.

RussellLuo: Personally, I think `csgclaw model auth` is sufficient. There's no need to add "login" unless there's a "logout" action.

xxx7xxxx: Maybe we will add it in the future, so be it.

cli/model/model.go
RussellLuo, line 1: auth/auth.go/auth_test.go => model/model.go/model_test.go?

internal/api/picoclaw.go
```diff
@@ -74,6 +80,7 @@ func (h *Handler) handlePicoClawEvents(w http.ResponseWriter, r *http.Request, b
 
 	_, _ = io.WriteString(w, ": connected\n\n")
 	flusher.Flush()
+	h.replayRecentPicoClawMessages(botID, r.Header.Get("Last-Event-ID"))
```
RussellLuo, line 95: Where does this `Last-Event-Id` come from?

xxx7xxxx: it’s expected from the PicoClaw SSE client / EventSource implementation. It's a standard header for SSE protocol.[1]

[1] https://http.dev/last-event-id

cli/model/model.go
xxx7xxxx: Done, plus adding shell completion subcommand.

internal/api/picoclaw.go
RussellLuo: OK, so this is partially done. PicoClaw will need to be adjusted later to match it.

RussellLuo:
LGTM (will merge it once the conflict's fixed).

xxx7xxxx:
Conflict fixed

xxx7xxxx:
@RussellLuo 

- I Restored the v0.2.5-style restart path which works after restarting server.
- Fixed the `csgclaw agent logs -f` side effect by reading the host-mounted `gateway.log` directly instead of holding the sandbox runtime lock, which will cause reconnecting failure
- Adds one-click “All members” selection when creating rooms and adding members to existing rooms.

xxx7xxxx:
@RussellLuo Fixed gateway.log mounting problem

internal/agent/box.go
```diff
@@ -114,11 +117,11 @@ type gatewayVolumeMount struct {
...
 }
 
-func gatewayVolumeMounts(hostWorkspaceRoot, projectsRoot string) []gatewayVolumeMount {
+func gatewayVolumeMounts(hostConfigRoot, projectsRoot string) []gatewayVolumeMount {
 	return []gatewayVolumeMount{
 		{
-			hostPath:  hostWorkspaceRoot,
-			guestPath: boxWorkspaceDir,
+			hostPath:  hostConfigRoot,
+			guestPath: boxPicoClawDir,
```
RussellLuo: Per the previous discussion, here we should not mount the hostConfigRoot anymore?

RussellLuo: Let's leave the mount for hostWorkspaceRoot as it is for now.

xxx7xxxx: corrected

internal/agent/box.go
```diff
@@ -114,19 +113,28 @@ type gatewayVolumeMount struct {
...
 	return []gatewayVolumeMount{
 		{
 			hostPath:  hostWorkspaceRoot,
 			guestPath: boxWorkspaceDir,
 		},
-		{
-			hostPath:  projectsRoot,
-			guestPath: boxProjectsDir,
-		},
 	}
```
RussellLuo, line 130: Why delete projectsRoot? This directory’s supposed to let all agents see the same shared projects folder, otherwise they can’t share build outputs at all!

xxx7xxxx: My bad, mis-deleting, reverted.

internal/agent/workspace.go
```diff
@@ -28,6 +28,9 @@ func ensureAgentWorkspace(agentName, template string) (string, error) {
 	if strings.TrimSpace(template) == "" {
 		return "", fmt.Errorf("workspace template is required")
 	}
+	if err := migrateLegacyAgentWorkspace(agentName, hostRoot); err != nil {
```
RussellLuo: Performing migration here doesn’t make sense:
1. Right now, only when we create an agent do we call `ensureAgentWorkspace()`, and at that point the agent usually doesn’t even exist (or it got deleted already);
2. Even if it does migrate successfully, the `copyEmbeddedWorkspace()` step is just gonna overwrite everything right after anyway.

xxx7xxxx: Good gotcha, fixed.

PR #30, opened 2026-05-07
Feat/feishu channel setup workflow

GatewayJ:


runtimes/picoclaw/manager/workspace/skills/feishu-channel-setup/SKILL.md
```diff
@@ -0,0 +1,335 @@
+---
+name: feishu-channel-setup
```
RussellLuo: Just call it `feishu`?

PR #38, opened 2026-05-08
Improve stored API key masking and reuse for model loading

xxx7xxxx:
**Summary**

- Fixes /api/v1/agent-profiles/models so model loading can reuse the stored agent API key when the request matches the same provider, base URL, and headers.

and

image

### DIFF
--- a/internal/agent/service_profiles.go
+++ b/internal/agent/service_profiles.go
@@ -3,7 +3,6 @@ package agent
 import (
 	"context"
 	"fmt"
-	"maps"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


internal/agent/service_profiles.go
```diff
@@ -161,6 +165,28 @@ func (s *Service) ListModelsForRequest(ctx context.Context, req ProfileModelRequ
...
+		return ""
+	}
+	stored := normalizeProfile(got.AgentProfile, got.Name, got.Description)
+	if stored.Provider != ProviderAPI || strings.TrimSpace(stored.APIKey) == "" {
+		return ""
+	}
+	if profile.BaseURL != stored.BaseURL {
+		return ""
+	}
+	if !maps.Equal(profile.Headers, stored.Headers) {
```
RussellLuo: Do the headers have to be the same? Is it sufficient if both the provider and base_url match?

xxx7xxxx: Good point. I relaxed this so stored API key reuse only requires the same agent, API provider, and base_url.

RussellLuo:
LGTM

PR #39, opened 2026-05-08
feat(sandbox): add Docker CLI provider and fix UI version label

wanghaojie124:
- Implement internal/sandbox/dockercli for docker run/inspect/exec lifecycle
- Register provider=docker with optional docker_cli_path in config
- Extend SandboxConfig validation and tests; update serve test for unknown providers
- Fix duplicate v in sidebar version when API returns git-describe v-prefixed string


RussellLuo:
LGTM

PR #47, opened 2026-05-11
[codex] fix upgrade prompt for git describe versions

youngbeom-shin:
## Summary

- Treat local git describe versions such as v0.2.9-4-gd1e3c28-dirty as their base release when checking for updates.
- Keep real prerelease versions such as v0.2.9-rc.1 eligible to update to the final release.
- Add regression coverage for both paths.

## Root Cause

The upgrade checker parsed git describe output as a semver prerelease. That made a local build based on v0.2.9 compare lower than the official v0.2.9 release, so the Web UI showed Update & Restart even when the release was already current.

## Validation

- env GOCACHE=/Users/shenyongfan/OpenCSG/csgclaw/.gocache go test ./internal/upgrade
- env GOCACHE=/Users/shenyongfan/OpenCSG/csgclaw/.gocache go test ./cli/upgrade
- env GOCACHE=/Users/shenyongfan/OpenCSG/csgclaw/.gocache go test ./internal/api -run 'TestHandleUpgrade|TestIMEventsIncludesUpgradeStatus'

### DIFF
--- a/internal/upgrade/upgrade.go
+++ b/internal/upgrade/upgrade.go
@@ -66,7 +66,8 @@ func (c Client) Check(ctx context.Context, currentVersion string) (CheckResult, 
 		CurrentVersion: normalizeSemver(currentVersion),
 		LatestVersion:  normalizeSemver(latestVersion),
 	}
-	if compareSemver(result.CurrentVersion, result.LatestVersion) >= 0 {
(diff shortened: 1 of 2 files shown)


RussellLuo:
The version upgrade flow has already been improved in another PR.

PR #56, opened 2026-05-12
Feishu skill

GatewayJ:


internal/channel/feishu/service.go
```diff
@@ -500,8 +504,8 @@ func defaultCreateChat(ctx context.Context, app AppConfig, req CreateChatRequest
 			Name(req.Title).
 			Description(req.Description).
 			OwnerId(req.CreatorID).
-			UserIdList(req.MemberIDs).
-			BotIdList([]string{}).
+			UserIdList([]string{}).
+			BotIdList(req.MemberAppIDs).
```
RussellLuo, line 515: At line 508 we're passing app IDs into `BotIdList`. Can you confirm this is expected to work with the Feishu API? The field name suggests it wants bot identifiers rather than app IDs, so it's worth double-checking that this doesn't silently break member creation for bot-based rooms.

internal/channel/feishu/service.go
```diff
@@ -451,14 +451,18 @@ func (s *Service) CreateRoom(req im.CreateRoomRequest) (im.Room, error) {
 		return im.Room{}, fmt.Errorf("feishu admin_open_id is required")
 	}
 	members := normalizeMembers(creatorID, req.MemberIDs)
-	memberIDs := members[1:]
+	memberBotIDs := members[1:]
```
RussellLuo, line 461: This updates the Feishu write path to translate member bot IDs into channel-specific IDs, but I don't see the corresponding reverse mapping on the read path. `ListRoomMembers`, `ListRoomMessages`, and uncached `ListRooms` still surface Feishu open IDs, so `csgclaw-cli room/member/message list --channel feishu` can still leak user-id semantics after this change. Should we translate those responses back to bot IDs here as well to keep the channel boundary consistent?

cli/bot/bot.go
```diff
@@ -73,7 +73,7 @@ func (c cmd) runList(ctx context.Context, run *command.Context, args []string, g
 	if err != nil {
 		return err
 	}
-	return command.RenderBots(globals.Output, run.Stdout, bots)
+	return command.RenderBotList(run.Program, globals.Output, run.Stdout, bots)
```
RussellLuo: This introduces program-specific branching inside the shared renderer layer. Would it be cleaner to keep `cli/command` focused on pure rendering and decide compact vs full output in `cli/bot/bot.go` based on `run.Program`? That would also make the compact path more explicit, since `RenderCompactBotList` currently relies on `RenderBotsTable` rather than rendering from `compactBotList(bots)`.

RussellLuo: Would it be simpler to drop `RenderBotList` and branch here in `cli/bot/bot.go` based on `run.Program`, then call `RenderCompactBotList` or `RenderFullBotList` directly? That keeps the program-specific decision in the command layer instead of pushing `csgclaw-cli` knowledge down into the shared renderer package.

cli/command/command.go
```diff
@@ -161,6 +162,35 @@ func RenderBots(output string, w io.Writer, bots []apitypes.Bot) error {
...
 }
 
+func RenderBotList(program string, output string, w io.Writer, bots []apitypes.Bot) error {
+	if strings.TrimSpace(program) == "csgclaw-cli" {
+		return RenderCompactBotList(output, w, bots)
+	}
+	return RenderFullBotList(output, w, bots)
+}
+
+func RenderCompactBotList(output string, w io.Writer, bots []apitypes.Bot) error {
```
RussellLuo, line 165: If this is intended to be the compact rendering path, should the `table` branch be driven from `compactBotList(bots)` as well? Right now the JSON branch is explicitly compact, but the table branch depends on `RenderBotsTable` staying aligned by coincidence rather than making the compact shape explicit here.

internal/channel/feishu/service.go
GatewayJ: Confirmed To invite the bot into the room, use app_id. An example value is something like cli_a10fbf7e94b8d01d.

cli/member/member.go
```diff
@@ -73,8 +73,8 @@ func (c cmd) runCreate(ctx context.Context, run *command.Context, args []string,
 	fs := run.NewFlagSet("member create", run.Program+" member create [flags]", "Add a member to a room.")
 	channelName := fs.String("channel", "csgclaw", "channel name: csgclaw or feishu")
 	roomID := fs.String("room-id", "", "target room id")
-	userID := fs.String("user-id", "", "user id to add")
-	inviterID := fs.String("inviter-id", "", "inviter user id")
+	userID := fs.String("user-id", "", "bot id to add")
```
RussellLuo: user-id => bot-id?

cli/member/member.go
```diff
@@ -83,7 +83,7 @@ func (c cmd) runCreate(ctx context.Context, run *command.Context, args []string,
 		return fmt.Errorf("member create does not accept positional arguments")
 	}
 	if *userID == "" {
-		return fmt.Errorf("user_id is required")
+		return fmt.Errorf("--user-id bot id is required")
```
RussellLuo: --user-id => --bot-id

cli/message/message.go
```diff
@@ -89,7 +89,7 @@ func (c cmd) runCreate(ctx context.Context, run *command.Context, args []string,
 		return fmt.Errorf("room_id is required")
 	}
 	if *senderID == "" {
-		return fmt.Errorf("sender_id is required")
+		return fmt.Errorf("--sender-id bot id is required")
```
RussellLuo: The field description for --sender-id already states that it is the bot ID, so there's no need to repeat "bot id" here.

PR #57, opened 2026-05-13
feat(notifier): in-server notifier delivery, IM fanout, Hub handlers, and notifier create fix

wanghaojie124:
Summary
Adds and hardens Notifier support on top of the latest upstream/main: webhook and remote-pull delivery via request_options.notifier, fanout to IM rooms, alignment with worker + notifier runtime semantics, Hub template workspace API handlers used by the bootstrap/publish flow, and a fix so creating a notifier no longer resolves the default sandbox worker Hub template (avoids runtime_kind mismatch with templates such as builtin/picoclaw-worker).

Changes
Notifier: relay/subscription wiring, webhook vs remote-pull modes, structured notify payloads into IM (with existing SSE/event paths as implemented in-tree).
Agent / IM / UI: notifier treated consistently as a worker-style agent with notifier runtime; profile/DM and related Web UI adjustments from this branch.
API: Hub template workspace detail and file handlers completed where callers expected them.
Agent service: skip default Hub worker template resolution when runtime_kind is notifier; regression test added.
Misc: minor web merge/style cleanup from the branch history.


internal/runtime/notifier/notify_card.go
```diff
@@ -0,0 +1,151 @@
...
+	maxNotifySummaryRunes   = 4000
+	maxNotifyMetaValueRunes = 2000
+	maxNotifyRawRunes       = 10000
+)
+
+// NotifyCard is the canonical JSON shape stored as IM message content for notifier deliveries.
+type NotifyCard struct {
+	Type          string       `json:"type"`
+	SchemaVersion int          `json:"schema_version"`
+	Provider      string       `json:"provider"`
```
RussellLuo: `csgclaw.notify_card` looks like a generic notification schema, but `Provider` is currently documented and encoded around `gitlab` / `github` / `generic`. That makes the contract narrower than the message type itself, and future providers would either need a schema change or be collapsed into `generic`. Consider defining `provider` as an open source identifier string, with GitHub/GitLab only as current built-in examples, and keeping the stable contract on the generic display fields (`title`, `subtitle`, `badge`, `summary`, `link`, `meta`, `raw`).

internal/agent/store.go
```diff
@@ -33,67 +34,104 @@ type legacyWorker struct {
...
+	CreatedAt         time.Time                `json:"created_at"`
+	Profile           string                   `json:"profile,omitempty"`
+	Provider          string                   `json:"provider,omitempty"`
+	ModelID           string                   `json:"model_id,omitempty"`
+	ReasoningEffort   string                   `json:"reasoning_effort,omitempty"`
+	AgentProfile      AgentProfile             `json:"agent_profile,omitempty"`
+	ProfileComplete   bool                     `json:"profile_complete"`
+	DetectionResults  []ProfileDetectionResult `json:"detection_results,omitempty"`
+	// LegacyDiskNotifierDetails is read from older state.json (top-level notifier_details); load migrates into agent.runtime_extensions.
+	LegacyDiskNotifierDetails map[string]any `json:"notifier_details,omitempty"`
```
RussellLuo: `LegacyDiskNotifierDetails` looks unnecessary now that notifier config has been moved to `runtime_extensions`. Consider removing this compatibility field and keeping a single persisted representation.

internal/server/http.go
```diff
@@ -53,6 +54,11 @@ func Run(opts Options) error {
 		ReadHeaderTimeout: 5 * time.Second,
 	}
 
+	if opts.Service != nil && handler != nil {
```
RussellLuo: Since notifier agents are expected to stay few in number, it may be cleaner to let each notifier own its own pull loop instead of running a global scanner from server startup. That keeps polling state and lifecycle local to the notifier runtime, and avoids leaking notifier-specific background orchestration into the server layer.

internal/bot/service.go
```diff
@@ -304,12 +305,12 @@ func (s *Service) createWorker(ctx context.Context, normalized CreateRequest) (B
...
 	created, ok := s.agents.Agent(workerAgentID(normalized))
 	if ok {
-		if strings.ToLower(strings.TrimSpace(created.Role)) != agent.RoleWorker {
+		if !strings.EqualFold(strings.TrimSpace(created.Role), agent.RoleWorker) {
 			return Bot{}, fmt.Errorf("agent id %q already exists with role %q", created.ID, created.Role)
 		}
 	} else {
-		var err error
-		created, err = s.agents.CreateWorker(ctx, agent.CreateAgentSpec{
+		var createErr error
```
RussellLuo: Using `createErr` here makes the block a bit noisier without really clarifying anything. Reusing the existing `err` would keep this path simpler.

internal/bot/service.go
```diff
@@ -489,6 +510,12 @@ func (s *Service) ensureChannelUser(ctx context.Context, channelName string, cre
 }
 
 func deriveAgentHandle(a agent.Agent) string {
+	if notifier.IsDeliveryWorker(a.Role, a.RuntimeKind) {
```
RussellLuo: This makes the bot layer aware of notifier-specific runtime behavior just to derive a handle. That coupling does not look necessary for the notifier feature itself; it seems cleaner to keep `deriveAgentHandle()` generic here.

internal/bot/store_test.go
```diff
@@ -227,3 +227,15 @@ func TestNewStoreRejectsInvalidState(t *testing.T) {
 		t.Fatalf("NewStore() error = %v, want role validation error", err)
 	}
 }
+
+func TestNewStoreRejectsLegacyNotifierBotRole(t *testing.T) {
```
RussellLuo: This test feels redundant with `TestNewStoreRejectsInvalidState`: `role=notifier` is just another invalid role now. If notifier-role compatibility is intentionally gone, adding a dedicated test for the legacy value seems unnecessary.

internal/server/http.go
```diff
@@ -33,6 +34,9 @@ type Options struct {
 	NoAuth      bool
 	Context     context.Context
 	OnReady     func()
+	// NotifierBackground starts notifier-specific background work after the API handler is wired.
+	// The implementation should block until ctx is cancelled; Run invokes it in its own goroutine.
+	NotifierBackground func(ctx context.Context, svc *agent.Service, deliver runtimenotifier.Fanouter)
```
RussellLuo: I don't think `NotifierBackground` should live in `internal/server`.

`internal/server` is a generic HTTP/server lifecycle layer, but this adds a notifier-specific startup hook to its public API. Since the pull supervisor is a long-running background loop, it also does not fit `OnReady`, which reads more like a one-shot readiness callback.

A cleaner split would be to keep `internal/server` focused on serving HTTP and start the notifier supervisor explicitly from the composition layer (`cli/serve` or runtime wiring), so the lifecycle ownership stays at the top level and notifier-specific wiring does not leak into the server package.

internal/bot/service.go
```diff
@@ -380,6 +381,25 @@ func agentProfileFromBotRequest(req apitypes.CreateAgentProfile) agent.AgentProf
 	}
 }
 
+func cloneRuntimeOptionsMapForBot(m map[string]any) map[string]any {
```
RussellLuo: This helper looks redundant. Since we already have `agentruntime.CloneAnyMap`, it would be cleaner to reuse that here instead of keeping a bot-local `cloneRuntimeOptionsMapForBot` wrapper.

internal/runtime/profile_extensions_policy.go
```diff
@@ -0,0 +1,161 @@
...
+	case KindNotifier:
+		return KindNotifier
+	default:
+		return strings.TrimSpace(kind)
+	}
+}
+
+// ProfileExtensionsPolicy defines how runtime_options (and related request_options) behave
+// for a concrete runtime_kind. Implementations register via RegisterProfileExtensionsPolicy.
+type ProfileExtensionsPolicy interface {
```
RussellLuo: ProfileExtensionsPolicy -> RuntimeOptionsPolicy

internal/runtime/profile_extensions_policy.go
```diff
@@ -0,0 +1,161 @@
...
+	FlatFromExtensionsMap(ext map[string]any) map[string]any
+	DetachPayloadFromProfileExtensions(profileExt map[string]any) map[string]any
+	StripNestedFromRequestOptions(ro map[string]any)
+	ProfileLLMFields(runtimeKind, baseURL, modelID string) (string, string)
+	// LooksLikeFlatStorageAtRoot detects notifier-shaped flat keys at the map root (used for legacy load heuristics).
+	LooksLikeFlatStorageAtRoot(m map[string]any) bool
+	ProfileComplete(isGatewayRuntime, llmComplete bool, agentExt map[string]any, runtimeKind string, mergedFlat map[string]any) bool
```
RussellLuo: IsComplete

internal/runtime/maps.go
```diff
@@ -0,0 +1,28 @@
+package runtime
```
RussellLuo: internal/runtime/maps.go -> internal/utils/maps.go

internal/api/handler.go
```diff
@@ -1561,6 +1595,20 @@ func parseRoomMembersPath(path string) (string, bool) {
...
+	}
+	return notifierbridge.NewFanout(h.im, func(roomID, senderID string, msg apitypes.Message, sender apitypes.User) {
+		h.publishMessageCreated(roomID, senderID, im.Message(msg))
+	})
+}
+
+// DeliverNotifierFanout posts notifier chat content (JSON notify card) to every IM room that includes this agent as a member and publishes SSE events.
+func (h *Handler) DeliverNotifierFanout(agentID string, content string) error {
+	return notifier.DeliverNotifierFanout(agentID, content, h.notifierFanoutBridge())
+}
```
RussellLuo: call API: codex SendMessage()

internal/agent/profile.go
```diff
@@ -112,6 +113,19 @@ func normalizeProfile(profile AgentProfile, fallbackName, fallbackDescription st
...
 
+// normalizeProfileForAgentRuntime applies profile normalization, then sets ProfileComplete using
+// runtime.ProfileCompleteFromAgentExtensions (gateway vs notifier vs other runtimes).
+// notifierFlat may be nil: completeness then uses ProfileExtensionsPolicyForKind(...).FlatFromExtensionsMap(agentRuntimeExt) when mergedFlat is empty.
+func normalizeProfileForAgentRuntime(profile AgentProfile, agentRuntimeExt map[string]any, fallbackName, fallbackDescription, runtimeKind string, notifierFlat map[string]any) AgentProfile {
+	out := normalizeProfile(profile, fallbackName, fallbackDescription)
+	llmComplete := profileIsComplete(out)
+	rk := normalizeRuntimeKind(runtimeKind)
+	pol := agentruntime.ProfileExtensionsPolicyForKind(agentruntime.NormalizeRuntimeKind(rk))
+	out.ProfileComplete = pol.ProfileComplete(isGatewayRuntimeKind(rk), llmComplete, agentRuntimeExt, rk, notifierFlat)
```
RussellLuo: 1. Simplify ProfileComplete(), only needs runtimOptions.
2. out.ProfileComplete = llmComplete && pol.ProfileComplete()

internal/runtime/profile_extensions_policy.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (defaultProfileExtensionsPolicy) LooksLikeFlatStorageAtRoot(map[string]any) bool {
+	return false
+}
+
+func (defaultProfileExtensionsPolicy) ProfileComplete(isGatewayRuntime, llmComplete bool, _ map[string]any, _ string, _ map[string]any) bool {
+	if isGatewayRuntime {
+		return llmComplete
+	}
+	return llmComplete
```
RussellLuo: Always return `true`.

internal/agent/service_profiles.go
```diff
@@ -320,61 +341,85 @@ func (s *Service) persistRecreatedAgent(ctx context.Context, id string, info age
...
-	}
-	if strings.TrimSpace(profile.Provider) == "" && strings.TrimSpace(req.Profile) != "" {
-		if _, cfg, err := s.llm.Resolve(req.Profile); err == nil {
-			profile = profileFromConfigModel(req.Name, req.Description, cfg)
-		} else if provider, modelID, ok := splitProfileSelector(req.Profile); ok {
-			profile.Provider = provider
-			if strings.TrimSpace(profile.ModelID) == "" {
-				profile.ModelID = modelID
-			}
+func (s *Service) profileForCreateRequest(ctx context.Context, spec *CreateAgentSpec) (AgentProfile, error) {
```
RussellLuo, line 337: ?

cli/serve/serve.go
```diff
@@ -479,32 +479,36 @@ func startServerWithConfigPath(ctx context.Context, run *command.Context, cfg co
...
-						fmt.Fprintln(run.Stdout, "Open this URL in your browser after startup.")
-					} else {
-						fmt.Fprintln(run.Stdout, "Opened this URL in your browser.")
-					}
-				}()
-			}
+	}
+	handler := server.NewHandler(serverOpts)
+	serverOpts.Handler = handler
+	go runtimewiring.RunNotifierPullSupervisor(ctx, svc, handler)
```
RussellLuo: put `PullSupervisor` related goroutine in OnReady().

PR #72, opened 2026-05-18
Support Responses API

xxx7xxxx:
- **Responses API support:** Adds Codex-facing `/llm/responses` routes and keeps Codex runtime on `wire_api = "responses"`.
- **Compatibility fallback:** Keeps upstream compatibility by falling back from unsupported `/responses` providers to chat completions behind the bridge.
- **Fix stale API provider state(nit):** Refreshes Codex profile/provider handling so changed API provider settings are probed/restarted correctly and stale failed probe/default state does not stick.

No architecture adjusted.

### DIFF
--- a/internal/llm/service.go
+++ b/internal/llm/service.go
@@ -368,6 +368,10 @@ func responsesCapabilityKey(profile agent.AgentProfile, baseURL string) string {
 }
 
 func responsesPayloadToChatPayload(payload map[string]any) (map[string]any, error) {
+	if responsesPayloadHasToolSemantics(payload) {
+		return nil, fmt.Errorf("active tool-use Responses requests are not supported by chat-completions fallback")
+	}
+
 	messages := make([]map[string]any, 0, 4)
 	if instructions := strings.TrimSpace(stringValue(payload["instructions"])); instructions != "" {
 		messages = append(messages, map[string]any{"role": "system", "content": instructions})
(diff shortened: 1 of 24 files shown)


internal/llm/service.go
```diff
@@ -149,6 +232,574 @@ func (s *Service) forwardRemoteChat(ctx context.Context, profile agent.AgentProf
...
+		s.responsesUnsupported = make(map[string]struct{})
+	}
+	s.responsesUnsupported[key] = struct{}{}
+}
+
+func responsesCapabilityKey(profile agent.AgentProfile, baseURL string) string {
+	return strings.TrimSpace(profile.Provider) + "\x00" + strings.TrimRight(strings.TrimSpace(baseURL), "/") + "\x00" + strings.TrimSpace(profile.ModelID)
+}
+
+func responsesPayloadToChatPayload(payload map[string]any) (map[string]any, error) {
```
RussellLuo, line 370: The chat-completions fallback only converts text messages and later re-encodes the result as `output_text`. It does not carry `tools`, tool outputs, or tool-call deltas through either the request or the streamed response path, so providers without native `/responses` will lose Codex tool use even though the request still appears to succeed. Please either map tool semantics through the fallback or reject tool-bearing requests explicitly.

xxx7xxxx: For this comment:

Updated fallback to allow text-only Responses requests that merely advertise tools, while still rejecting active tool-use states so chat fallback does not silently drop tool semantics.

For wire_api:

Removed the generic wire_api plumbing from the agent/runtime profile boundary and kept wire_api = "responses" owned by the Codex runtime config generation, since it is currently a Codex-specific implementation detail.

PR #77, opened 2026-05-19
feat(hub): add remote registry store and default opencsg templates

wanghaojie124:
Enable listing and fetching templates from remote hub registries (e.g. csgclaw.opencsg.com), merge builtin/local/opencsg defaults into hub config, tolerate partial registry list failures with warnings, and cap remote JSON responses. Update serve, API workspace handling, and agent-related tests.


internal/config/config.go
```diff
@@ -118,22 +118,49 @@ func (c HubConfig) Resolved() HubConfig {
 		return c
 	}
 
-	out := make([]HubRegistryConfig, 0, len(c.Registries))
+	user := make([]HubRegistryConfig, 0, len(c.Registries))
```
RussellLuo: `user` reads a bit too user-facing here. `configured` seems more accurate for values coming from config.

internal/config/config.go
```diff
@@ -118,22 +118,49 @@ func (c HubConfig) Resolved() HubConfig {
...
+	registry.Name = strings.TrimSpace(registry.Name)
+	registry.Kind = strings.TrimSpace(registry.Kind)
+	registry.Path = strings.TrimSpace(registry.Path)
+	registry.URL = strings.TrimSpace(strings.TrimRight(registry.URL, "/"))
+	registry.Token = strings.TrimSpace(registry.Token)
+	return registry
+}
+
+func mergeHubRegistries(defaults, user []HubRegistryConfig) []HubRegistryConfig {
+	byName := make(map[string]HubRegistryConfig, len(user))
```
RussellLuo: Since this map is built from configured registries, `configuredByName` may be clearer than `byName`.

internal/config/config.go
```diff
@@ -244,6 +281,8 @@ const (
 	BoxLiteProvider                 = "boxlite"
 	DefaultHubRegistry              = "builtin"
 	DefaultHubPublishRegistry       = "local"
+	DefaultOfficialHubRegistryName  = "opencsg"
```
RussellLuo: Would `official` be a better default registry name here than `opencsg`? The URL can still point to `https://csgclaw.opencsg.com`, while the registry name stays role-based and UI-friendly.

RussellLuo (commented):
If we rename the default remote registry to `official`, the Web UI should also localize that source tag. Please update `web/static/app.js`, especially `localizeTemplateSourceTag`, so it shows `official` in English and `官方` in Chinese.

PR #85, opened 2026-05-20
[codex] Refine hub template detail layout

youngbeom-shin:
## Summary

- Apply the hub template detail layout refinement from PR #73.
- Keep the current role badge while grouping role, runtime, and source badges into one responsive row.
- Remove the duplicated description block and make the inspector grid use balanced responsive columns in both Vite source and legacy static assets.

## Validation

- PATH=/Users/shenyongfan/.nvm/versions/node/v22.13.0/bin:$PATH pnpm --dir web/app typecheck
- make test was attempted, but the local sandbox blocks test HTTP listeners (bind: operation not permitted) and the environment is missing boxlite for existing serve tests.


youngbeom-shin:
Closing in favor of a renamed branch and PR title that follow the requested naming convention.

PR #91, opened 2026-05-20
refactor(notification): channel notification bots with pull delivery and web UI

wanghaojie124:
Move notifier delivery from runtime agents to type=notification channel bots, with relay pull/push, dedicated API routes, and supervisor fixes that read stored tokens for pull eligibility. Web UI adds notification sidebar section, simplified notifier profile controls, and CLI bot list limited to normal bots.


internal/bot/service.go
```diff
@@ -85,6 +85,9 @@ func (s *Service) List(channel, role string) ([]Bot, error) {
 
 	filtered := make([]Bot, 0, len(all))
 	for _, b := range all {
+		if IsNotificationBot(b) {
```
RussellLuo: This unconditional skip means the normal list API now hides notification bots for every channel. For the CLI behavior we want to avoid adding a separate `bot notification list` command: `bot list --channel csgclaw` should include both `type=normal` and `type=notification`, while `bot list --channel feishu` should continue to return only normal bots. Please move that channel-specific filtering into this list path and drop the separate notification-list surface/tests.

internal/api/router.go
```diff
@@ -148,7 +151,10 @@ func (h *Handler) registerChannelRoutes(router chi.Router) {
 			r.Post("/", h.createBot)
 		})
 		r.Route("/{channel}/bots/{id}", func(r chi.Router) {
+			r.Get("/", h.handleBotByID)
+			r.Patch("/", h.handleBotByID)
 			r.Delete("/", h.deleteBot)
+			r.Post("/notifications", h.pushNotificationBot)
```
RussellLuo: Route structure suggestion: keep common bot CRUD on the generic `/api/v1/channels/{channel}/bots` and `/api/v1/channels/{channel}/bots/{id}` routes, and keep channel-specific actions on explicit channel routes only. In practice that means generic routes handle `GET/POST /bots` and `GET/PATCH/DELETE /bots/{id}`, `/api/v1/channels/csgclaw/bots/{id}/notifications` stays csgclaw-only, and `/api/v1/channels/feishu/bots/{id}/events` stays feishu-only. This avoids duplicating the csgclaw/feishu CRUD routes while also avoiding a generic `/{channel}/bots/{id}/notifications` endpoint for channels that do not support notifications.

internal/bot/model_type_test.go
```diff
@@ -0,0 +1,27 @@
+package bot
+
+import "testing"
+
+func TestNormalizeBotTypeDefaultsToNormal(t *testing.T) {
```
RussellLuo: These tests are small and cover `model.go` behavior directly; consider moving them into `internal/bot/model_test.go` instead of keeping a separate file just for bot type.

internal/api/handler.go
```diff
@@ -395,6 +404,15 @@ func (h *Handler) handleBots(w http.ResponseWriter, r *http.Request) {
 
 	switch r.Method {
 	case http.MethodGet:
+		if r.URL.Query().Get("type") == bot.BotTypeNotification {
+			bots, err := h.botSvc.ListNotificationBots(channelName)
```
RussellLuo: Prefer keeping this type-specific filtering inside `internal/bot`: the handler can pass `type=notification` as list criteria, and the service can decide whether to return normal, notification, or both for each channel.

cli/bot/bot.go
```diff
@@ -70,10 +74,44 @@ func (c cmd) runList(ctx context.Context, run *command.Context, args []string, g
...
+func (c cmd) runNotificationList(ctx context.Context, run *command.Context, args []string, globals command.GlobalOptions) error {
+	fs := run.NewFlagSet("bot notification list", run.Program+" bot notification list [flags]", "List notification bots.")
+	channelName := fs.String("channel", "csgclaw", "channel name: csgclaw")
+	if err := fs.Parse(args); err != nil {
+		return err
+	}
+	if len(fs.Args()) != 0 {
+		return fmt.Errorf("bot notification list does not accept positional arguments")
+	}
+	bots, err := run.APIClient(globals).ListNotificationBots(ctx, *channelName)
```
RussellLuo: This should probably go through the same bot list path with a `type` filter instead of a notification-specific list API; that keeps CLI/API behavior aligned and avoids a separate notification list surface.

internal/config/advertise_base_url_test.go
```diff
@@ -0,0 +1,37 @@
+package config
+
+import "testing"
+
+func TestAdvertiseBaseURLForClient(t *testing.T) {
```
RussellLuo: This is a small config helper test; consider moving it into `internal/config/config_test.go` instead of adding a separate test file for `advertise_base_url`.

internal/config/config.go
```diff
@@ -316,6 +316,14 @@ func ListenPort(listenAddr string) string {
 	return port
 }
 
+// AdvertiseBaseURLForClient returns the base URL exposed to Web UI and external webhook callers.
+func AdvertiseBaseURLForClient(server ServerConfig) string {
```
RussellLuo: Consider renaming this to `ResolveAdvertiseBaseURL`: the helper does more than return a client value, since it resolves `server.advertise_base_url` with a fallback derived from `listen_addr`.

internal/api/handler.go
```diff
@@ -922,6 +982,7 @@ func presentHubTemplate(item hub.Template) apitypes.HubTemplate {
 		Role:        item.Role,
 		RuntimeKind: item.RuntimeKind,
 		Image:       item.Image,
+		Env:         item.Env,
```
RussellLuo: This added `Env` assignment does not compile at the current PR head: `apitypes.HubTemplate` has no `Env` field, and `hub.Template` has no `Env` field either. `go test ./cli ./internal/api ./internal/bot ./internal/apiclient` fails before the API and CLI packages can build. Please either add the field to both template types or remove this assignment.

internal/bot/service.go
```diff
@@ -85,6 +85,9 @@ func (s *Service) List(channel, role string) ([]Bot, error) {
 
 	filtered := make([]Bot, 0, len(all))
 	for _, b := range all {
+		if !shouldIncludeBotInList(b, normalizedChannel, botType) {
```
RussellLuo, line 88: The store-backed bots are filtered by `botType`, but configured Feishu bots are appended afterward without seeing that filter. As a result, `/api/v1/channels/feishu/bots?type=notification` can still return configured Feishu bots, even though those are normal bots and notification bots are meant to be excluded for Feishu. Consider skipping this append when `botType` normalizes to `notification`, or pass the type filter into `appendConfiguredFeishuBots`.

RussellLuo:
LGTM

PR #94, opened 2026-05-21
fix(web): support sandbox deployment routing

jialudev:
## Summary

- Resolve frontend assets, favicon, shared worker, and API requests relative to the current mount point so the web UI can run behind sandbox path prefixes.
- Switch the SPA to hash routing to avoid server-side route fallback assumptions in sandbox deployments.
- Use registry/name Hub template routes instead of encoded slash IDs so template detail and workspace file requests survive proxy and sandbox routing layers.

## Testing

- Not run by Codex in this PR creation step; user reports this branch was tested locally.

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -866,8 +866,11 @@ func (h *Handler) handleHubTemplates(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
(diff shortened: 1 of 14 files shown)


RussellLuo:
LGTM

PR #97, opened 2026-05-21
Support Threads

xxx7xxxx:
**Summary**

- Adds Matrix-shaped IM thread support across storage, API routes, bot compatibility, and Codex/PicoClaw session isolation.
- A room-grouped Threads tab

**Screenshots**
image

image


internal/im/service.go
```diff
@@ -612,12 +653,20 @@ func (s *Service) Bootstrap() Bootstrap {
...
+}
+
+func (s *Service) ListRoomsWithOptions(opts ListMessagesOptions) []Room {
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 
 	rooms := make([]Room, 0, len(s.rooms))
 	for _, room := range s.rooms {
-		rooms = append(rooms, s.presentRoomLocked(*room))
+		presented := s.presentRoomLocked(*room)
```
RussellLuo, line 665: `presentRoomLocked(*room)` clones the full `room.Messages` slice here, so the default `ListRooms()` response still includes thread replies. That contradicts the new timeline contract and leaks replies through `/api/v1/rooms` and `/api/v1/im/bootstrap` even when `include_thread_replies` is not requested. Please filter through `presentMessagesLocked(..., false)` on the default path and add a regression test for the room/bootstrap endpoints.

internal/im/service.go
```diff
@@ -343,6 +383,7 @@ func savePersistedRooms(statePath string, rooms []Room) ([]persistedRoom, error)
 			IsDirect:    room.IsDirect,
 			Members:     append([]string(nil), room.Members...),
 			Messages:    relativePath,
+			Threads:     cloneThreadStates(room.Threads),
```
RussellLuo, line 386: `room.Threads` is now persisted here, but `DeleteUser()` still only rewrites `room.Messages`. That means thread snapshots can keep deleted users’ content in `ThreadState.Context`, and deleted-root threads can remain orphaned in `state.json`. Please prune or rebuild `room.Threads` from the surviving top-level messages during user deletion, and add a regression test for that path.

RussellLuo (commented):
This change introduces a fairly substantial threads design across IM storage, API shape, bot bridges, Codex session isolation, and UI behavior, but most of that design is currently implicit in the code. It would be very helpful to add a more complete design note to this PR describing the intended model and key decisions: how threads are represented and persisted, why context is snapshotted, how thread replies differ from the main timeline, how bot/Codex session isolation works, and what the expected invariants are. Having that design context attached to the PR would make the change much easier to review now and much easier to maintain later.

RussellLuo (commented):
From the information architecture perspective, I think Threads would likely work better as a secondary entry under Messages rather than as a top-level peer of Messages / Agents / Hub. Threads still seem to be a message-domain view, similar to Rooms and DMs, instead of an independent workspace category. Keeping it at the top level makes the navigation hierarchy feel a bit inconsistent, because it mixes message containers and message-derived views in the same layer. Unless Threads are intentionally being positioned as a cross-conversation workbench that users will enter directly and manage independently, I would suggest nesting them under Messages as a second-level entry or view switch.

xxx7xxxx:
## Enhanced design summary: CSGClaw IM threads

This PR adds an incremental, Matrix-shaped thread model to the existing CSGClaw IM APIs rather than replacing the IM stack with full Matrix Client-Server semantics.

### Data model and persistence
- A thread is anchored by its root message ID inside a room or DM.
- Thread replies use `relates_to: { rel_type: "m.thread", event_id: <root_message_id> }`, matching the Matrix relation shape while keeping CSGClaw's current room/auth/session model.
- `ThreadState` is persisted per room for explicitly started threads. Existing JSONL messages remain backward-compatible because relation and thread summary fields are optional.
- Thread roots expose a computed summary: reply count, latest reply, participants, current-user participation, and deterministic context summary.

### Timeline behavior and invariants
- Thread replies stay out of the main room/DM timeline by default, Slack-style.
- `/api/v1/messages?include_thread_replies=true` is the explicit opt-in path for full message history including thread replies.
- Room/bootstrap views now preserve the same default timeline contract and do not expose thread replies implicitly.
- Thread roots must be top-level messages in the same room; nested thread roots and missing/cross-room roots are rejected.
- Deleting a user now rebuilds thread state from surviving messages, so deleted users' content does not remain in thread context snapshots and deleted-root threads are pruned.

### Thread context for agents
- Starting a thread snapshots bounded hidden context around the root message: up to 5 top-level messages before, the root, and up to 2 top-level messages after, subject to payload bounds.
- The snapshot is for agent/LLM understanding only; it is not inserted as visible thread messages.
- v1 uses deterministic summaries instead of an LLM summarizer: root excerpt plus context counts.

### Bot, PicoClaw, and Codex bridge behavior
- Bot events include `thread_root_id` and hidden `thread_context` so adapters can preserve topic/thread isolation.
- Bot send accepts optional `thread_root_id`, allowing replies to stay in the same thread.
- Codex bridge conversation/session identity is scoped by `room_id` for top-level room work and `room_id:thread_root_id` for thread work, preventing thread prompts/tools from bleeding into room-level sessions.
- Tool-call result messages are attached to the relevant response thread to reduce room-level noise while preserving traceability.

### UI behavior
- Message rows expose a hover-only “Reply in thread” affordance, similar to Slack.
- Threads open in a right-side thread panel with its own composer and automatic scroll-to-latest behavior.
- The sidebar includes a Threads view grouped by room/DM so users can scan active thread conversations without mixing replies into the main timeline.
- The UI also keeps thread previews compact, strips code-fence prefixes from preview text, prevents horizontal timeline scroll, and shows running-status dots for agent DMs.

### Navigation note
I kept Threads as a dedicated sidebar tab for this v1 because it is meant to act as a cross-room thread inbox grouped by room/DM, while still being scoped to the message domain. I agree it should not become a separate protocol/workspace surface; if we later add second-level navigation under Messages, this view can move there without changing the storage/API model.

internal/im/service.go
xxx7xxxx: Fixed in the rebased PR. The default room/bootstrap presentation now preserves the same thread filtering contract as the top-level message timeline: thread replies are not exposed unless an endpoint explicitly opts into thread replies. I also added regression coverage in `TestHandleThreadRoutesAndMessageFiltering` for both `GET /api/v1/rooms` and `GET /api/v1/bootstrap`, so this does not regress through the room/bootstrap surfaces.

internal/im/service.go
xxx7xxxx: Fixed. `DeleteUser()` now rebuilds persisted thread state after pruning messages, which removes deleted-root threads and regenerates context snapshots from surviving messages only while preserving the original thread creation time. I added `TestDeleteUserRebuildsThreadStateFromSurvivingMessages` to cover deleted context messages, deleted replies, and deleted-root thread pruning.

PR #98, opened 2026-05-21
add Codex tool permission approval UI

GatewayJ:
## Summary

- Project Codex tool calls and permission requests as structured agent activity messages.
- Add an in-memory Codex permission broker, decision API, and Web UI permission cards.
- Update tool-call filtering and message rendering so structured tool activity can be hidden while pending permissions remain visible.

## Why

Codex tool calls were previously rendered as plain text, and ACP permission requests were automatically allowed by the backend. This PR makes tool activity machine-readable in the IM timeline and lets users approve or reject Codex permission requests from the Web UI.

## Validation

- `go test ./internal/runtime/codex ./internal/channel/codexbridge ./internal/api`
- `pnpm --dir web/app test`

Note: pnpm reported the local Node version warning (`wanted >=22.13.0 <25`, current `v26.0.0`), but the test and typecheck commands completed successfully.


internal/api/router.go
```diff
@@ -45,6 +45,8 @@ func (h *Handler) registerCoreRoutes(router chi.Router) {
 			r.Get("/status", h.handleCLIProxyAuthStatus)
 			r.Post("/login", h.handleCLIProxyAuthLogin)
 		})
+		r.Post("/runtime/permissions/{request_id}/decision", h.handleRuntimePermissionDecision)
+		r.Post("/codex/permissions/{request_id}/decision", h.handleCodexPermissionDecision)
```
RussellLuo: If this PR is standardizing permission decisions under the runtime API, keeping a second `codex/permissions` endpoint preserves the old abstraction boundary and leaves two public routes for the same operation. Unless there is an external compatibility requirement, I would drop this route and keep `runtime/permissions` as the only API surface.

cli/serve/serve.go
```diff
@@ -829,6 +831,38 @@ type codexBridgeManager interface {
...
+		return nil
+	}
+	decider := withPermissions.PermissionDecider()
+	if decider == nil {
+		return nil
+	}
+	return codexPermissionActivityDecider{permission: decider}
+}
+
+type codexPermissionActivityDecider struct {
```
RussellLuo: This adapter feels a bit too high in the stack for `cli/serve.go`. It is already carrying Codex-specific activity decision semantics rather than pure server wiring, so I would move it down next to the Codex runtime implementation (for example under `internal/runtime/codex`) and have `serve.go` only construct or reference it. That keeps the CLI/bootstrap layer thin now, and still leaves room to generalize later if another agent runtime needs the same activity interface.

PR #99, opened 2026-05-22
docs: add Agentfile spec v1

RussellLuo:


docs/hub/agentfile_spec.md
```diff
@@ -0,0 +1,1186 @@
...
+- 不应包含真实 secret。
+- 建议只使用小写字母、数字、点号、短横线和下划线。
+
+设计理由：
+
+- `name` 更适合作为展示名称，不适合作为稳定身份。
+- 同步、升级、去重、收藏、权限控制、来源追踪和引用都需要稳定 ID。
+- `updated_at` 不能替代模板 ID。
+
+#### `version`
```
wanghaojie124, line 208: Suggestion: template versioning in repos and pinning old versions

The spec defines version semantics well, but not how multiple versions coexist in a repo/registry or how users select an older one.

Please clarify:

Storage model — overwrite templates/<slug>/, versioned subdirs (<slug>/<version>/), or manifest-only with remote artifacts?
User refs — does registry/slug support a version (@0.1.0 / /0.1.0)? What is “latest” when omitted?
Agent lifecycle — are created agents pinned to the template revision, or do they follow future bumps?
A short subsection with one layout example + one CLI pinning example would close the gap with the existing id + version upgrade guidance.

PR #100, opened 2026-05-22
Align PicoClaw runtime workspace layout

jialudev:
Align the PicoClaw sandbox creation flow with OpenClaw by using a consistent runtime-root layout, mount structure, and gateway logging behavior across both runtimes.

### DIFF
--- a/internal/agent/workspace.go
+++ b/internal/agent/workspace.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"csgclaw/internal/runtime/picoclawsandbox"
 	"csgclaw/internal/templates"
 )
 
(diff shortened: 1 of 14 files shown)


internal/agent/workspace.go
```diff
@@ -70,7 +71,7 @@ func agentWorkspaceRoot(agentName string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return filepath.Join(agentHome, hostWorkspaceDir), nil
+	return picoclawsandbox.WorkspaceRoot(agentHome), nil
```
RussellLuo: This helper now hard-codes the PicoClaw workspace layout, but it is still used from runtime-agnostic paths such as `HubPublishSpec()`.

That means a non-PicoClaw agent, especially OpenClaw, can resolve its exported workspace snapshot from the wrong root: `.picoclaw/workspace` instead of `.openclaw/workspace`.

The template-overlay path is still correct because each runtime provisions its own workspace root, but this shared helper no longer looks safe as a generic agent-layer API.

Could we either:

1. make this helper runtime-aware, for example resolve by `got.RuntimeKind`, or
2. remove the default workspace-root assumption from `agent/workspace.go` and let each runtime or caller provide the host workspace root explicitly?

jialudev: Optimized.

PR #117, opened 2026-05-26
Feat(threads): Support Threads of PicoClaw

xxx7xxxx:
- Adds PicoClaw thread support
- Fixes some minor bugs in UI: cant't @ in threads, the replies will be showed in main room instead of threads and so on.


xxx7xxxx:
Related to https://github.com/OpenCSGs/picoclaw/pull/10

PR #119, opened 2026-05-26
feat(FormControls): add data-invalid attribute

youngbeom-shin:
- Add data-invalid attribute to label
- Import inputs.css in globals.css
- Add portal-error-500 color token
- Add error-focus-ring variable


jialudev:
LGTM

Can you take a screenshot of the adjusted input CSS and take a look? All subsequent inputs are extracted into components for comparison to avoid overwriting your style optimization
@youngbeom-shin

PR #125, opened 2026-05-27
fix(web): use secondary close button in upgrade modal

youngbeom-shin:
- use secondary close button in upgrade modal


youngbeom-shin:
截屏2026-05-27 14 16 16

PR #133, opened 2026-05-28
fix(web): refine message bubble layout

jialudev:
- Adapt message bubble colors to the global theme and make compact bubbles size to their content.


jialudev:
image

PR #137, opened 2026-05-29
fix(web): hide upgrade controls for csghub sandbox

GatewayJ:
- Add `[server].show_upgrade` to control whether the Web UI shows upgrade actions, defaulting to enabled.
- Set `show_upgrade = false` for managed deployments that cannot self-upgrade.


web/app/src/hooks/workspace/useWorkspaceController.ts
```diff
@@ -261,6 +261,7 @@ export function useWorkspaceController() {
       currentUserID: displayData.current_user_id,
       usersById: conversation.usersById,
       collapsedWorkspaceGroups,
+      hideUpgradeControls: String(bootstrapConfig?.sandbox_provider ?? "").trim().toLowerCase() === "csghub",
```
RussellLuo: This hard-codes upgrade visibility to `sandbox_provider == "csghub"`.

That couples a user-facing product decision to the runtime backend, so CSGHub users can never opt in and non-CSGHub users can never opt out. Consider exposing a dedicated `config.toml` flag instead, for example `[server].show_upgrade = true` by default, and let the UI follow that setting.

internal/config/config.go
```diff
@@ -30,6 +30,7 @@ type ServerConfig struct {
 	AdvertiseBaseURL string
 	AccessToken      string
 	NoAuth           bool
+	HideUpgrade      bool
```
RussellLuo: The public config key and API field are both `show_upgrade`, but the internal Go field is `HideUpgrade`, so the meaning flips when reading call sites like `!cfg.Server.HideUpgrade`.

This works, but it adds avoidable cognitive overhead. Consider renaming the internal field to `ShowUpgrade` as well, so the config, API, frontend, and Go model all use the same positive direction.

RussellLuo:
What does the UI look like when upgrade actions are hidden? For example, does the settings menu still show a version section with only the current version, or is the whole version-and-updates area removed? A screenshot in the PR would help confirm the intended UX.

GatewayJ:
> What does the UI look like when upgrade actions are hidden? For example, does the settings menu still show a version section with only the current version, or is the whole version-and-updates area removed? A screenshot in the PR would help confirm the intended UX.

截屏2026-05-29 下午1 49 30

internal/config/config.go
```diff
@@ -30,6 +30,11 @@ type ServerConfig struct {
 	AdvertiseBaseURL string
 	AccessToken      string
 	NoAuth           bool
+	ShowUpgrade      *bool
```
RussellLuo: Why does this need to be a pointer bool? Would a plain bool be enough here?

PR #140, opened 2026-05-29
feat(agent): add avatar field to agent model

youngbeom-shin:
- Add Avatar field to Agent struct
- Update CreateAgentSpec to include Avatar
- Modify UpdateRequest to handle Avatar
- Implement Avatar handling in service methods
- Update API responses to include Avatar
- Add Avatar support in bot and agent creation
- Enhance tests for Avatar functionality
- Update UI components to display Avatar


internal/agent/service.go
```diff
@@ -1393,6 +1428,48 @@ func isRetryableRemoveAllError(err error) bool {
 	return strings.Contains(lower, "directory not empty") || strings.Contains(lower, "permission denied")
 }
 
+func safeSandboxNameForAgent(id, name string) string {
```
RussellLuo: This introduces an auto-generated sandbox identifier derived from `name/id/hash`, but we plan to add a dedicated agent `id` field later and validate its legality and uniqueness there. Given that direction, I do not think this PR should change sandbox naming in this way. It would be better to keep this part unchanged for now and align it with the later id-based model.

PR #141, opened 2026-05-29
fix(web): compact upgrade version menu

GatewayJ:
- Show a compact version row in the settings menu, with update actions only when an update is available.
- Keep local builds and hidden upgrade controls focused on the current version only.


GatewayJ:
有版本更新：
d92379bf-e54f-452d-9ffe-27e1ee896ce3

无版本更新
51d5a2c3-6163-4faf-a239-3127d4b3238a

PR #155, opened 2026-06-01
Refactor(architecture): Use Participants API/Models

xxx7xxxx:
- Refactor points
  - Introduce `Participant` as the channel-visible identity model, replacing public bot/user surfaces with `/api/v1/channels/{channel}/participants` APIs, CLI commands, and docs.
  - Rewire agent/runtime provisioning around participant IDs, including manager bootstrap recreation, PicoClaw/OpenClaw config/env, LLM bridge routes, and channel delivery paths.
  - Update the Web UI, embedded agent skills, API clients, and focused tests to use participant contracts while reducing legacy bot-command/API coverage.
  - Dependent PR: https://github.com/OpenCSGs/picoclaw/pull/12
- Test points:
  - All added cli like `csgclaw participant/pt`
  - Runtime: Codex, PicoClaw (without adapting/testing OpenClaw)
  - Channel CSGClaw (without adapting/testing FeiShu)


GatewayJ:
CLI 没有足够定义，会有新命令么。新旧命令的迁移对照方案？

xxx7xxxx:
@GatewayJ updated for cli

xxx7xxxx:
And this PR is only for co-review so I've made it draft. Next step will be implementing Recreating Agent warning for outdated image with inheriting skills.

xxx7xxxx:
@RussellLuo, ready to review.

cli/csgclawcli/app.go
```diff
@@ -68,7 +68,8 @@ func (a *App) AddCommand(commands ...command.Command) {
 
 func (a *App) registerDefaultCommands() {
 	a.AddCommand(
-		bot.NewCmd(),
+		participantcmd.NewCmd(),
```
RussellLuo, line 71: This removes the existing `bot config` CLI entrypoint without introducing an equivalent replacement, even though `/api/v1/channels/feishu/config` is still supported server-side. That means existing Feishu configuration workflows will break on upgrade. If the plan is to move this under participant or channel semantics, it would be safer to add the replacement command first and only remove `bot config` after the migration path exists.

xxx7xxxx: Added csgclaw participant/pt config to replace bot config

RussellLuo:
Follow-up idea for a future PR: after the participant migration settles, it may be worth fully retiring `internal/bot` and moving the remaining notification/compat responsibilities into the participant path as well. Right now the bot package still appears to carry notification delivery/config glue and some compatibility behavior, so the system is in a mixed bot/participant state. That is understandable for this transition, but finishing that consolidation later would reduce duplicate concepts and make the ownership boundaries much clearer.

PR #156, opened 2026-06-02
feat: add slash skill invocation

GatewayJ:
- Added canonical slash-command XML normalization via shared slashcommand module, and standardized message flow to store/relay XML across Web/API/IM while converting to /slash only at Feishu egress.
- Added boundary-based normalization: CSGClaw normalizes on message creation; Feishu normalizes / shorthand on ingress.
- Fixed thread composer regressions using contentEditable parsing for slash/mention tokens (placeholder/suggestion matching, token-boundary replacement, and unified delete behavior) and added regression coverage.
- Minor readability refactor for bot event text selection early-return logic.


RussellLuo (commented):
The current design puts slash-command expansion in the local `handleCreateMessage` path, which makes the behavior channel-specific. `csgclaw` messages created here can populate `AgentContent`, but Feishu and future channels have separate create-message paths and will not automatically reuse this logic. Consider introducing a shared `slashcommand` module instead of a skill-specific `skillinvocation` module, and run that from a channel-agnostic message ingress/enrichment layer. That keeps slash parsing and expansion consistent across channels and avoids adding more handler-specific API and model special cases as more slash commands are introduced.

RussellLuo (commented):
Moving from a skill-specific `skillinvocation` path to a shared `slashcommand` module is a clear improvement. The remaining issue is where normalization happens: right now canonical slash-command formatting is enforced in the API handlers, but `csgclaw` and `feishu` still have separate channel send-message implementations, and only `csgclaw` eventually reaches `im.CreateMessage`. That means the internal message invariant still depends on which ingress path was used. It would be more robust to keep `slashcommand` as the shared module, but invoke it from the actual channel message-creation boundary (`channel/csgclaw` and `channel/feishu`), or another shared helper reused by both channel services, rather than only from the HTTP layer. That would make slash-command handling a true channel-level invariant instead of an API-specific one.

PR #160, opened 2026-06-03
feat(web): add workspace file preview dialog

jialudev:
- Adds a workspace file preview dialog with Markdown preview and source modes.


jialudev:
image

PR #161, opened 2026-06-03
Feat: Detect oudated images and Recreate enhancement

xxx7xxxx:
- Detect outdated default runtime images and recreate agents/managers with the latest default template image.
- Preserve user-installed skills during recreate while refreshing built-in template skills from the latest embedded workspace.
- Add local Docker/BoxLite runtime image listing and update the manager rebuild UI to prefer defaults, include local images, and handle long image refs.


xxx7xxxx:
Upgrade button added.

image
image

internal/agent/workspace.go
```diff
@@ -119,6 +119,142 @@ func overlayWorkspaceTree(srcRoot, dstRoot string) error {
...
+		return nil, nil, nil
+	}
+	if err != nil {
+		return nil, nil, fmt.Errorf("stat workspace skills: %w", err)
+	}
+	if !info.IsDir() {
+		return nil, nil, nil
+	}
+
+	tempDir, err := os.MkdirTemp("", "csgclaw-preserve-skills-*")
```
RussellLuo, line 147: This uses the OS temp directory for skill preservation. In our production Docker sandbox setup the runtime workspace is mapped onto NAS, so copying skills to an unmanaged local temp dir can break recreate if the host temp space is isolated, cleaned up, or not under the same storage controls as the workspace. Please keep the preservation dir inside a controlled csgclaw workspace/state location instead of `os.MkdirTemp("")`.

RussellLuo:
`ListImages` is now implemented in both `dockercli` and `boxlitecli`, but the dispatch still happens in the API layer by branching on the sandbox provider. That leaks provider-specific knowledge upward and makes `internal/api` responsible for wiring runtime capabilities, which feels like the wrong layer.

I think this would be cleaner if `ListImages` were modeled as part of the sandbox abstraction and implemented behind the sandbox interface, so callers can ask the sandbox for local images without knowing whether the backend is Docker or BoxLite.

That would keep the API layer provider-agnostic, and it would also make future sandbox providers easier to add because this code would not need another provider-specific branch.

internal/api/router.go
```diff
@@ -49,6 +50,7 @@ func (h *Handler) registerCoreRoutes(router chi.Router) {
 		})
 		r.Post("/agent-profiles/models", h.handleAgentProfileModels)
 		r.Get("/agent-profile-defaults", h.handleAgentProfileDefaults)
```
RussellLuo, line 52: `/runtime/images` feels too generic for what this endpoint actually returns. The payload is not really “runtime information”; it is a candidate list used by agent create/recreate flows. A more explicit route like `/api/v1/agents/image-candidates` would make the intent clearer: these are image options for agent operations, not a general runtime resource.

xxx7xxxx: Fixed along with the ListImages comment.

PR #168, opened 2026-06-03
feat(im): add room message clearing

GatewayJ:
- Add channel-scoped clearMessages API for CSGClaw rooms
- Add Web UI confirmation flow for clearing room messages


internal/im/service.go
```diff
@@ -876,6 +876,27 @@ func (s *Service) DeleteConversation(conversationID string) error {
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	room, ok := s.rooms[roomID]
+	if !ok {
+		return Room{}, fmt.Errorf("room not found")
+	}
+	room.Messages = []Message{}
+	room.Threads = []ThreadState{}
+	if err := s.saveLocked(); err != nil {
```
RussellLuo: This change only mutates a single room (`Messages` and `Threads`), but it persists through `saveLocked() -> SaveBootstrap()`, which rewrites every room's session JSONL and then scans/cleans the whole `sessions` tree. That makes the I/O cost much larger than the scope of the operation.

`ClearRoomMessages()` would be a good candidate for a room-scoped persistence path instead: extract the single-room save logic from `savePersistedRooms()`, reuse the existing JSONL/blob encoding and cleanup helpers, update only `sessions/<room>.jsonl` and `sessions/blobs/<room>/`, and still refresh this room's persisted metadata in `state.json` (especially `Threads`).

That would preserve the current on-disk format and correctness, while avoiding a full rewrite of all rooms for a single-room clear.

internal/im/service.go
```diff
@@ -876,6 +876,27 @@ func (s *Service) DeleteConversation(conversationID string) error {
 	return s.DeleteRoom(conversationID)
 }
 
+func (s *Service) ClearRoomMessages(roomID string) (Room, error) {
```
RussellLuo, line 908: `ClearRoomMessages()` would be a good place to emit a `room.messages_cleared` event after the clear has been persisted successfully.

This operation is a domain-level state transition, not just an HTTP-side effect. Emitting the event here keeps "clear + persist + notify" tied to the same operation, so every caller of `im.Service.ClearRoomMessages()` gets the same synchronization behavior.

That also gives the frontend an authoritative signal to update local state for this room, instead of relying only on request-local cleanup logic.

web/app/src/hooks/workspace/useConversationController.ts
```diff
@@ -755,6 +771,32 @@ export function useConversationController({
...
+    const threadRootIDsToClear = new Set((roomBeforeClear?.threads || []).map((thread) => thread.root_message_id || ""));
+
+    try {
+      await clearRoomMessagesRequest(roomID);
+    } catch (err) {
+      setComposerError(localizeError(err.message, t));
+      return;
+    }
+
+    setBootstrapData((current) =>
```
RussellLuo: Now that the backend returns a cleared room and also emits `room.messages_cleared`, the frontend can use the same authoritative room update path for both the initiating tab and other tabs.

Using request-local ID filtering here keeps the current page on a heuristic cleanup path, while other tabs converge through the event payload. It would be more robust to upsert the cleared room from the HTTP response immediately, and let `room.messages_cleared` apply the same room-level update everywhere else.

That keeps all clients converging through the same state shape and avoids race windows where the initiating page temporarily differs from the server truth.

internal/api/router.go
```diff
@@ -137,6 +137,7 @@ func (h *Handler) registerChannelRoutes(router chi.Router) {
 		r.Route("/csgclaw/rooms", func(r chi.Router) {
 			r.Get("/", h.listRooms)
 			r.Post("/", h.createRoom)
+			r.Post("/{id}:clearMessages", h.clearCsgclawRoomMessages)
```
RussellLuo: This operation looks more like an IM-native room capability than a cross-channel room capability. It only applies to the local IM model, persists through `im.Service`, and is not meaningful for channels like Feishu.

Because of that, `/api/v1/rooms/...` may be a cleaner home than `/api/v1/channels/csgclaw/rooms/...`: the route would match the underlying ownership of the state, and `handleClearRoomMessages` could call `h.im.ClearRoomMessages(roomID)` directly without going through `requireLocalChannel()` / `csgclaw.Service` just to get back to the IM service.

That would make the API boundary a bit clearer: IM-native room operations under `/api/v1/rooms`, and only genuinely cross-channel room operations under `/api/v1/channels/{channel}/rooms`.

cli/room/room.go
```diff
@@ -38,6 +38,8 @@ func (c cmd) Run(ctx context.Context, run *command.Context, args []string, globa
 		return c.runList(ctx, run, args[1:], globals)
 	case "create":
 		return c.runCreate(ctx, run, args[1:], globals)
+	case "clear-messages":
```
RussellLuo: For the same reason as the route discussion, this capability currently looks IM-native rather than cross-channel. If the API eventually belongs under `/api/v1/rooms/...` instead of the channel-scoped room surface, it may be better not to add `room clear-messages` to the cross-channel CLI yet, since that would start to solidify it as a generic room/channel capability.

Deferring the CLI surface until the API ownership is settled could keep the command model a bit cleaner.

PR #182, opened 2026-06-08
feat(web): add config settings with save and restart

wanghaojie124:
- Expose common config.toml fields in the Web UI settings modal, add structured config API endpoints with bootstrap validation before save, and restart the daemon or prompt for manual restart after saving.


RussellLuo:
I think the API shape here would be clearer if we group these endpoints under `server` and keep the responsibilities explicit.

A cleaner split would be:
- `GET /api/v1/server/config`
- `PUT /api/v1/server/config`
- `POST /api/v1/server/restart`
- `GET /api/v1/server/restart/status`

The reason is that `PUT /server/config` is about persisting configuration, while restart is a separate lifecycle action on the running service. Keeping restart outside the config resource makes that boundary easier to understand.

Also, I think the current `apply` naming is a bit misleading. In the current flow, `PUT` already saves the config, and the extra step is really restarting the service so the saved config takes effect. Naming that operation as `restart` would match the actual behavior more closely.

So if the goal is a complete and consistent server-facing API surface for this feature, `server/config + server/restart + server/restart/status` seems like the cleanest option.

PR #186, opened 2026-06-09
Fix: Restore Upgrade button & Migrate legacy admin state to participnt records

xxx7xxxx:
- Restores the worker Upgrade action in agent list and detail views, including coverage for restart-required and hidden-lifecycle-action states.
- Migrates the local admin identity from legacy `u-admin` to participant/user `admin` across IM bootstrap, persisted participant records, room members, senders, mentions, and session state.
- Bootstraps `admin` as a `human` participant with no backing agent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,10 @@ dist/
 *.test
 *.out
 
+# Python cache artifacts
+__pycache__/
+*.py[cod]
+
 # Web frontend local artifacts
 node_modules/
 web/app/playwright-report/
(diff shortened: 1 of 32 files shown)


internal/templates/embed/openclaw-manager/workspace/skills/manager-worker-dispatch/scripts/__pycache__/manager_worker_api.cpython-314.pyc
RussellLuo, line 1: `__pycache__` should be removed

xxx7xxxx: Removed and updated .gitignore

PR #188, opened 2026-06-09
refactor(embed): drive picoclaw image tags from agent.toml version

wanghaojie124:
- Embed templates are embedded directly without dist/ staging. Local build-all bumps version and image.ref; GitLab CI on main builds and pushes images only when version changes since the previous commit.


.gitlab/ci-picoclaw-version-check.sh
```diff
@@ -0,0 +1,148 @@
...
+read_agent_toml_image_ref_at_ref() {
+  read_agent_toml_field_at_ref "$1" "$2" "ref"
+}
+
+image_ref_tag() {
+  local ref="$1"
+  printf '%s' "${ref##*:}"
+}
+
+find_previous_main_commit() {
```
RussellLuo: This baseline is too narrow for main-branch push pipelines. Comparing only `HEAD` vs `HEAD~1` misses cases where one push contains multiple commits and the version bump happened in an earlier commit, so the final pipeline can incorrectly skip the image build.

I recommend diffing against the previous processed default-branch tip instead, ideally `CI_COMMIT_BEFORE_SHA` with `HEAD~1` only as a fallback. That makes the decision depend on the whole pushed range, not just the last parent.

A simple shape would be:
`compare_base=${CI_COMMIT_BEFORE_SHA:-$(git rev-parse HEAD~1)}`
then either compare `version` at `${compare_base}` vs `HEAD`, or more directly check whether `internal/templates/embed/<name>/agent.toml` changed in `${compare_base}..HEAD` and, if it did, always validate that `version` matches the `[image].ref` tag before deciding whether to build.`

xxx7xxxx:
Can we add a development document to indicate which content and files should be updated if Picoclaw/Openclaw has been upgraded for local dev?

for example, I have updated some code in local PicoClaw,  and I made the local dev image with tag like 2026.6.9, how should I update stuff to use that image.

wanghaojie124:
> Can we add a development document to indicate which content and files should be updated if Picoclaw/Openclaw has been upgraded for local dev?
> 
> for example, I have updated some code in local PicoClaw, and I made the local dev image with tag like 2026.6.9, how should I update stuff to use that image.

Thanks for the suggestion.

For local PicoClaw dev, update only the embed template Dockerfiles (picoclaw-manager / picoclaw-worker) and set ARG PICOCLAW_IMAGE to your local base image tag (e.g. .../picoclaw:2026.6.9). Then run make build-all. No Makefile or CI changes needed.

PR #193, opened 2026-06-10
fix(agent): apply gateway profile on save and fix glm token limits

wanghaojie124:
- Recreate picoclaw sandbox when profile inputs change, bootstrap manager after first complete profile, and normalize completion tokens above the 32768 thinking budget. 
- Refresh agent UI state after save and show online indicator when the manager runtime is running.

### DIFF
--- a/web/app/src/hooks/workspace/useAgentController.ts
+++ b/web/app/src/hooks/workspace/useAgentController.ts
@@ -95,6 +95,9 @@ type AgentWithProfile = {
   profile?: AgentProfileLike | null;
 };
 
+const AGENT_RUNTIME_SYNC_INTERVAL_MS = 2_000;
+const AGENT_RUNTIME_SYNC_TIMEOUT_MS = 120_000;
+
 export function useAgentController({
   activeConversationId,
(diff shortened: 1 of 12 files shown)


web/app/src/hooks/workspace/useAgentController.ts
```diff
@@ -516,6 +520,8 @@ export function useAgentController({
       setManagerProfileData(saved);
       setProfileDraft({ ...profileToDraft(saved), agent_id: MANAGER_AGENT_ID });
       await refreshManagerProfile();
+      await syncAgentStateUntilRunning(MANAGER_AGENT_ID);
```
RussellLuo: This now waits for the manager to become `running` after every profile save, but the backend change here does not guarantee that every successful save will also start the runtime. For cases like a config-only gateway update or a manager that intentionally stays stopped, the UI can sit in a busy state until the 120s timeout even though the save already succeeded. Consider only waiting when the save path actually triggers bootstrap/start/recreate, or treating `stopped` as an acceptable terminal state for this poll.

PR #200, opened 2026-06-11
feat(team): expose agent display names in task APIs and UI

jialudev:
- Show friendly agent display names instead of raw IDs in team task APIs and the Tasks page, e.g. assigned to data-2-worker rather than agent-ymkx7q.


jialudev:
image

PR #202, opened 2026-06-11
fix(web): restore manager profile save after avatar auto-save

wanghaojie124:
- Stop avatar edits from submitting incomplete LLM profiles, validate required model and API fields before save, and sync manager runtime in the background with clearer incomplete-setup feedback.


web/app/src/hooks/workspace/useAgentController.ts
```diff
@@ -830,11 +834,15 @@ export function useAgentController({
...
   }
 
-  async function saveAgentPage(draftOverride?: AgentDraft): Promise<void> {
-    const draftToSave = draftOverride ?? agentPageDraft;
+  async function saveAgentPage(): Promise<void> {
+    const draftToSave = agentPageDraft;
     if (!draftToSave || !selectedAgentForPage?.id) {
       return;
     }
+    if (agentProfilePageSaveDisabled(draftToSave, selectedAgentForPage)) {
```
RussellLuo: This top-level completeness guard now blocks avatar/description-only saves for incomplete agents as well. The `meta-only` branch below can persist those fields without sending `agent_profile`, so users in `profile_incomplete` can no longer save safe metadata edits. Could we defer the completeness check until after `hasProfileOrRuntimeChange` and only gate saves that actually submit profile/runtime changes?

PR #213, opened 2026-06-16
feat(web): improve room member management and tooltip styles

youngbeom-shin:
Summary:\n- Improve room member management flows and unify tooltip styling/behavior across the app.\n- Add member removal UX, refactor header actions, and align button/icon sizing with design references.\n\nValidation:\n- make build\n- local server restarted on 18080\n\nNotes:\n- Includes new log.svg icon and tooltip global styling updates.


jialudev:
LGTM @RussellLuo

jialudev (commented):
LGTM

internal/api/router.go
```diff
@@ -90,6 +90,7 @@ func (h *Handler) registerCoreRoutes(router chi.Router) {
 				r.Route("/members", func(r chi.Router) {
 					r.Get("/", h.listRoomMembers)
 					r.Post("/", h.addRoomMembers)
+					r.Post("/remove", h.handleRoomMembersRemovePath)
```
RussellLuo: This removal flow hides the target member IDs in the request body behind POST /rooms/{id}/members/remove. If this endpoint is intentionally batch-oriented, a collection custom method such as POST /api/v1/rooms/{room_id}/members:batchDelete would make that contract explicit and align better with the existing custom-method style; otherwise, for single-member removal, DELETE /api/v1/rooms/{room_id}/members/{member_id} would be clearer.

internal/api/handler.go
```diff
@@ -1434,6 +1434,15 @@ func (h *Handler) handleRoomMembersByID(w http.ResponseWriter, r *http.Request,
 	writeJSON(w, http.StatusOK, members)
 }
```
RussellLuo, line 1441: This wrapper only extracts the room ID from the path and immediately forwards to handleRemoveRoomMembersRequest, but there is no second entry point reusing the shared helper here. Consider folding these into a single handler to avoid an extra indirection unless another remove-members route is expected to share the implementation.

PR #214, opened 2026-06-17
feat(web): add floating manager chat and split conversation components

jialudev:
- Add a floating chat shortcut in the bottom-right corner for quick access to the manager conversation.
- Extract conversation UI pieces into reusable business components under `components/business/ConversationPane` and compose them in both the full conversation page and the floating chat panel.


jialudev:
image
image

PR #215, opened 2026-06-17
fix(upgrade): require official bundle marker for auto updates

wanghaojie124:
- add release bundle marker validation before automatic install replacement
- show manual upgrade state in the UI for non-official installs
- point Windows installer launchers at the bundled executable


web/app/src/models/upgradeStatus.ts
```diff
@@ -24,7 +26,17 @@ export function formatSidebarVersionLabel(version: unknown): string {
...
-  return raw.length > 0 && raw !== "dev" && raw.endsWith("+local");
+  return (
+    raw.length > 0 &&
+    (raw === "dev" || raw.endsWith("+local") || raw.endsWith("-dirty") || /-\d+-g[0-9a-f]+/i.test(raw))
+  );
+}
+
+export function isLocalBuildUpgradeStatus(status: UpgradeStatus | null | undefined, version: unknown): boolean {
+  return (
+    isLocalBuildVersion(version) ||
```
RussellLuo, line 37: This helper currently treats any unsupported install with a non-empty reason as a local build. With the new serve-time gate, `not_official_bundle` should normally be rejected before the UI loads, so folding all unsupported reasons into `local build` makes the state model harder to reason about. Consider checking `reason == "local_build"` explicitly and keeping other unsupported reasons separate, even if they are currently unreachable in the normal runtime flow.

PR #235, opened 2026-06-24
feat(skillhub): add default system skills

GatewayJ:
## Summary
- add an embedded system skill source for default skills
- expose built-in skills in Skill Hub and prevent deleting or uploading over them
- install default system skills during agent create, start, and recreate flows
- move skill-installer out of runtime templates into the shared system skill source

## Validation
- go test ./internal/api ./internal/skillhub ./internal/agent ./internal/templates
- go test ./...
- pnpm --dir web/app check
- git diff --check
- make build


GatewayJ:
ae5b538b-2807-453d-a1db-c72827a91f0d

internal/agent/service.go
```diff
@@ -1816,18 +1816,28 @@ func (s *Service) provisionRuntime(ctx context.Context, rt agentruntime.Runtime,
...
+	}
+	if err := s.installDefaultSystemSkills(req.AgentName, runtimeKind, role); err != nil {
+		return fmt.Errorf("install default system skills: %w", err)
+	}
+	return nil
+}
+
 func (s *Service) provisionRuntimeForAgent(ctx context.Context, rt agentruntime.Runtime, got Agent, workspaceOverlay string) error {
 	if s == nil || rt == nil {
 		return nil
```
RussellLuo, line 1839: Now that `provisionRuntimeWithDefaultSystemSkills` takes a `role`, do we still need to distinguish roles here? At the moment the default system skill set still looks fixed to `skill-installer`, so it may be simpler to keep this path role-agnostic until we actually have manager/worker-specific defaults.

internal/agent/service.go
```diff
@@ -1816,18 +1816,28 @@ func (s *Service) provisionRuntime(ctx context.Context, rt agentruntime.Runtime,
 	return provisioner.Provision(ctx, req)
 }
 
+func (s *Service) provisionRuntimeWithDefaultSystemSkills(ctx context.Context, rt agentruntime.Runtime, runtimeKind string, req agentruntime.ProvisionRequest, role string) error {
```
RussellLuo: Now that `provisionRuntimeWithDefaultSystemSkills` takes a `role`, do we still need to distinguish roles here? At the moment the default system skill set still looks fixed to `skill-installer`, so it may be simpler to keep this path role-agnostic until we actually have manager/worker-specific defaults.

internal/agent/service.go
```diff
@@ -554,7 +554,7 @@ func (s *Service) ensureManager(ctx context.Context, forceRecreate bool, imageOv
 		if err != nil {
 			return err
 		}
-		if err := s.provisionRuntime(ctx, runtimeImpl, runtimeKind, agentruntime.ProvisionRequest{
+		if err := s.provisionRuntimeWithDefaultSystemSkills(ctx, runtimeImpl, runtimeKind, agentruntime.ProvisionRequest{
```
RussellLuo: I’d suggest keeping the name `provisionRuntime` for simplicity. `provisionRuntimeWithDefaultSystemSkills` exposes too much implementation detail, and the name should ideally remain stable as more logic is added later.

PR #239, opened 2026-06-24
feat(model): Add OpenCSG Builtin model

xxx7xxxx:
Also updated the OpenCSG sign-in flow so the new tab still opens OpenCSG login, but it no longer asks the callback to redirect that tab back to CSGClaw, which causes redundant CSGClaw tabs.

image


internal/agent/model_providers.go
```diff
@@ -9,15 +9,18 @@ import (
...
 	"csgclaw/internal/modelprovider"
 )
 
 const (
+	ModelProviderIDOpenCSG    = "opencsg"
 	ModelProviderIDCSGHubLite = "csghub-lite"
 	ModelProviderIDCodex      = ProviderCodex
 	ModelProviderIDClaude     = ProviderClaudeCode
 
+	ModelProviderKindOpenCSG          = ProviderCSGHub
```
RussellLuo: ProviderCSGHub => ProviderOpenCSG?

xxx7xxxx: Sure, then I will change all CSGhub text to OpenCSG? which @jialudev has written before.

image

xxx7xxxx: Updated CSGHub to OpenCSG

PR #244, opened 2026-06-25
fix(config): preserve Windows hub registry paths

wanghaojie124:


internal/config/config.go
```diff
@@ -924,7 +924,11 @@ func parseBoolValue(raw string) (bool, error) {
 }
 
 func parseRawStringValue(raw string) string {
-	return strings.Trim(strings.TrimSpace(raw), `"`)
+	raw = strings.TrimSpace(raw)
```
RussellLuo, line 927: Since this now relies on `strconv.Unquote` rather than a plain trim, it would help to add a short comment here explaining why: this preserves canonical TOML-escaped content for values like Windows hub registry paths, and avoids double-escaping on later save cycles. For example: `// Preserve escaped TOML string content such as Windows paths. A plain trim would cause double-escaping on later save cycles.`

wanghaojie124: Thanks for the suggestion. I think we can keep this helper without an inline comment for now: the implementation is still fairly direct, and the new tests already capture the main reason for the change, namely preserving escaped TOML content such as Windows paths and avoiding double-escaping on save. Adding that context here may end up encoding a very specific bug history into a general-purpose parsing helper. If we find this pattern becoming non-obvious elsewhere, I would be more inclined to document it at the config parsing level instead.

PR #246, opened 2026-06-26
feat(hub): lazy load official template workspaces

wanghaojie124:
- Switch the official template registry to the OpenCSG hub API and add staging override support.
- Load hub template workspace directories and file previews on demand instead of during template detail fetches.

### DIFF
--- a/internal/template/remote_store.go
+++ b/internal/template/remote_store.go
@@ -1,95 +1,433 @@
 package template
 
 import (
-	"archive/tar"
-	"bytes"
-	"compress/gzip"
 	"context"
+	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"time"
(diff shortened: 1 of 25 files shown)


internal/template/remote_store.go
```diff
@@ -1,97 +1,433 @@
...
-	for _, item := range payload {
-		items = append(items, templateFromAPI(item))
+
+	items := make([]Template, 0, len(payload.Data))
+	for _, repository := range payload.Data {
+		id, err := normalizeRemoteTemplateID(repository.Path)
+		if err != nil {
+			return nil, fmt.Errorf("invalid remote hub template path %q: %w", repository.Path, err)
+		}
+		item, err := s.getTemplate(ctx, id, repository)
```
RussellLuo, line 113: This turns any single bad repository or invalid `agent.toml` into a failure for the entire `official` registry listing, because `List()` now returns immediately on the first per-repo error. That means one non-template repo or one broken manifest can make the whole Hub template list disappear. Consider logging and skipping invalid repositories here so the registry can still return partial results.

PR #250, opened 2026-06-26
fix(config): migrate legacy official hub URL

wanghaojie124:


internal/config/config.go
```diff
@@ -132,10 +132,18 @@ func normalizeHubRegistry(registry HubRegistryConfig) HubRegistryConfig {
 	registry.Kind = strings.TrimSpace(registry.Kind)
 	registry.Path = strings.TrimSpace(registry.Path)
 	registry.URL = strings.TrimSpace(strings.TrimRight(registry.URL, "/"))
```
RussellLuo, line 134: This now rewrites every `kind = "remote"` registry to `DefaultOfficialHubRegistryURL`, not just the legacy official endpoint. That means custom remotes like the existing `team` registry and `${HUB_URL}`-backed configs get overwritten on load/save, and `csgclaw serve` will persist that change because `NeedsMigrationRewrite()` becomes true. Please scope the migration to the official registry or the legacy official URL only, and keep arbitrary remote registries intact.

RussellLuo (commented):
One follow-up: `config.toml` already supports the local hub `path` and remote hub `url`, but those fields are still not exposed in the UI settings. It would be better to show and allow editing them there as well, so the UI stays aligned with the actual config surface instead of leaving part of the hub setup as file-only configuration.

PR #267, opened 2026-06-30
feat(tasks): support agent and team task workflows

jialudev:
- Supports assigning tasks to single Agents and Teams through shared taskcore APIs and CLI/HTTP flows.
- Keeps Team metadata in ~/.csgclaw/state.json under teams.items[] while moving task data into ~/.csgclaw/tasks/.
- Storage shape: state.json -> teams.items[]; tasks/index.json -> task root index; tasks/<root-task-id>/{root.json,children.json,events.jsonl,approvals.json,presence.json} -> task tree, lifecycle, approvals, and worker presence.
- Relationship shape: Agent -> taskcore root task; Team -> TeamMeta in state.json plus taskcore roots; parent task -> child tasks -> assigned Agent participants -> execution room lifecycle.
- Adds realtime workspace updates so newly created Agents, Rooms, Participants, and task-related room changes appear without manual refresh.
- Improves dispatch, claim, and completion flows, including compact task assignment events and cleaner task board names.


jialudev:
image

image

internal/agenttask/service.go
```diff
@@ -0,0 +1,298 @@
...
+	if agentID == "" {
+		return taskcore.Task{}, fmt.Errorf("agent_id is required")
+	}
+	title := strings.TrimSpace(input.Title)
+	if title == "" {
+		return taskcore.Task{}, fmt.Errorf("title is required")
+	}
+
+	name := s.agentDisplayName(agentID)
+	user, room, err := s.im.EnsureAgentUser(im.EnsureAgentUserRequest{
```
RussellLuo, line 77: `CreateAgentTask` canonicalizes the requested agent ID and then calls `EnsureAgentUser`, but it never verifies that the agent exists in `agent.Service`. A typo such as `--agent-id u-fronend` can create a task and direct room for a phantom worker instead of failing early. Consider checking `s.agents.Agent(agentID)` when the agent service is available and returning a not-found error before creating IM state.

RussellLuo:
The current split between `taskcore` and `agenttask` makes sense to me: one is the reusable task domain/core, and the other is the agent-specific application layer around IM/direct-room orchestration.

What feels a bit inconsistent is mainly the package layout/naming. If we want a cleaner structure later, I would lean toward grouping them under a common `internal/task/...` namespace, for example `internal/task/core` and `internal/task/agent`, while still keeping them as separate submodules rather than collapsing everything into one `task` package.

So this is more of a follow-up structure suggestion than a blocker for this PR.

jialudev:
image

internal/agenttask/service.go
jialudev: Fixed

PR #272, opened 2026-07-02
fix(agent): unify worker runtime selection compatibility

wanghaojie124:
- Unify worker runtime selection by introducing runtime_name + sandbox_enabled and adding backend compatibility for legacy runtime_kind payloads.
- Update agent create/update API, CLI, and field mask handling so split runtime fields stay consistent and older clients do not break.
- Fix worker default runtime and image selection, including sandbox fallback when Codex CLI is unavailable and default image resolution for sandbox workers.
- Refresh the Workspace AgentProfileModal runtime UX with Codex CLI / sandbox selection, helper copy, and frontend runtime normalization updates.
- Add backend and frontend regression coverage for runtime normalization, legacy compatibility, default image fallback, modal switching behavior, and UI contract expectations.


internal/api/handler.go
```diff
@@ -469,6 +492,62 @@ func bootstrapConfigView(ctx context.Context, cfg config.Config, hubSvc *hub.Ser
...
+		Name:           name,
+		Role:           agent.RoleWorker,
+		RuntimeName:    agent.RuntimeNameCodex,
+		SandboxEnabled: false,
+		RuntimeKind:    agent.RuntimeKindCodex,
+	}
+	if h == nil || h.svc == nil {
+		return spec
+	}
+	if _, err := h.svc.Runtime(agent.RuntimeKindCodex); err == nil {
```
RussellLuo: `defaultWorkerCreateSpec()` still treats Codex as available when `h.svc.Runtime(agent.RuntimeKindCodex)` succeeds, but actual worker creation now gates local runtime on `codexcli.Locator().Locate()` in `CreateWorker()`. Those checks are not equivalent. On a machine where the Codex runtime is registered but the Codex CLI is not installed, bootstrap already marks Codex unavailable, yet the auto-created worker here will still default to Codex and later fail with `codex cli not installed`. Consider reusing the same availability check as `workerRuntimeChoices()` so this fallback stays consistent.

PR #279, opened 2026-07-02
fix(skillhub): disable installed remote skill installs

GatewayJ:
## Summary
- disable remote skill install actions when the skill already exists locally
- share remote-path skill name normalization between the API adapter and skill hub model helpers
- keep local skills fresh when the remote install list is opened or refreshed
- stabilize the remote install button width across install, installing, and installed states

## Validation
- `pnpm --dir web/app exec vitest run tests/models/skillhub.test.ts tests/api/skills.test.ts tests/hooks/useWorkspaceHubSelection.test.tsx tests/components/WorkspaceTabPanels.test.tsx -t "skill|AgenticHub|remote|official|extracts|matches|disables remote skill install actions|shows remote skills|refreshes local skills"`
- `pnpm --dir web/app lint` (0 errors; existing warnings remain)
- touched-file Prettier check
- `git diff --check`

## Impact
Remote skill rows now show installed local skills as disabled `Installed` buttons in both light and dark themes, preventing duplicate install attempts and avoiding layout shift while an install is in progress.


GatewayJ:
image

web/app/src/models/skillhub.ts
```diff
@@ -61,6 +61,29 @@ export function hasSkillName(
   return (skills || []).some((item) => normalizeSkillName(item?.name) === value);
 }
 
+export function hasInstalledRemoteSkill(
+  installedSkills: readonly SkillSummary[] | null | undefined,
+  remoteSkill: SkillSummary | null | undefined,
+): boolean {
+  const installedName = remoteSkillInstalledName(remoteSkill);
+  return hasSkillName(installedSkills, installedName);
```
RussellLuo: Matching only by the local skill name seems too strong here: a same-named local skill does not necessarily mean this exact remote skill was installed before. Unless we persist a stronger remote identity such as source/remotePath on install, it may be safer to treat this as a name collision warning and let the user decide.

GatewayJ:
image

PR #294, opened 2026-07-06
fix: harden Windows agent runtime handling

wanghaojie124:
- Default unset sandbox providers to Docker and update docs/tests for the new behavior.
- Fix Windows Codex runtime liveness checks, app-server shutdown, and delete retries for locked files.
- Share retryable RemoveAll handling across agent, skill, template, and migration cleanup paths.
- Require template-declared environment variables in the agent creation UI and backend save flow.
- Update frontend and Go tests for status mapping, required env validation, and Windows cleanup retries.


internal/fileutil/remove_test.go
```diff
@@ -0,0 +1,49 @@
+package fileutil
```
RussellLuo: Suggest moving this package into `internal/utils`.

PR #312, opened 2026-07-09
fix(web): use full language option labels

jialudev:
- Updates language controls to show full Simplified Chinese and English labels instead of abbreviated options.
- Adjusts segmented control sizing so the longer labels fit cleanly in the sidebar menu and settings page.


jialudev:
image

PR #318, opened 2026-07-10
fix(agent): stabilize codex agent setup flows

wanghaojie124:
- Fall back to /tmp when the hub workspace temp root disappears during template fetches.
- Retry transient Windows file-lock failures when rebuilding Codex agent runtime directories.
- Hide upgrade actions for Codex agents that do not have image metadata.
- Preserve Unicode task titles for manager agent-teams on Windows by avoiding shell-written task JSON for the default path.


RussellLuo:
LGTM

PR #319, opened 2026-07-10
fix(web): keep sidebar nav icons embedded

youngbeom-shin:
## Summary
- Render workspace sidebar navigation icons with existing React SVG components instead of CSS mask URLs.
- Remove the unused mask class for sidebar navigation icons.


jialudev:
The new React SVG path leaves each icon wrapped in an unstyled inline `span`. Because the navigation row uses a 24px line height, the inline SVG participates in baseline layout and can appear vertically shifted with different fonts, browsers, zoom levels, or device pixel ratios.

Rendering the icon directly as the flex item and adding `display: block` to `.primaryNavIcon svg` removes the baseline/descender space and keeps the icon and label vertically centered consistently.

jialudev (commented):
image


LGTM

PR #322, opened 2026-07-10
feat(web): clarify activity rows and conversation counts

jialudev:
- This change improves the agent activity timeline by using multi-key, normalized merging (`item_id`, `id`, `tool_call_id`) so related command lifecycle/output messages are shown as one clean activity row instead of fragmented entries.
- It also updates conversation headers to show visible/total counts when messages are filtered and adds dark-theme polish to activity row/segment/badge styling for better readability.


RussellLuo:
LGTM

PR #330, opened 2026-07-13
fix(web): refine message sidebar interactions

jialudev:
- Remove the redundant top-level room creation action while keeping room creation in the Rooms section.
- Align dark-mode sidebar selection with the primary navigation and keep group action tooltips anchored under UI scaling.


jialudev:
image

PR #345, opened 2026-07-15
fix(web): replace native title tooltips

youngbeom-shin:
## Summary

- Add a shared tooltip primitive for consistent custom tooltip rendering.
- Replace native DOM title tooltips across the web UI with the shared tooltip.
- Update tooltip-related tests to assert the new custom tooltip behavior.

### DIFF
--- a/web/app/src/components/ui/Button/Button.tsx
+++ b/web/app/src/components/ui/Button/Button.tsx
@@ -1,6 +1,7 @@
 import { forwardRef } from "react";
 import type { ButtonHTMLAttributes, ReactNode } from "react";
 import { classNames } from "@/shared/lib/classNames";
+import { Tooltip } from "../Tooltip";
 
 export type ButtonVariant =
   | "primary"
@@ -61,6 +62,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(function Button
     loading = false,
     loadingLabel,
     size = "md",
+    title,
     type = "button",
     variant = "secondaryGray",
     "aria-busy": ariaBusy,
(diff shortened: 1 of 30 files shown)


web/app/src/components/ui/Button/Button.tsx
```diff
@@ -99,6 +101,8 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(function Button
       ) : null}
     </button>
   );
+
+  return title ? <Tooltip content={title}>{button}</Tooltip> : button;
```
jialudev, line 105: Could we keep Button as a pure button primitive and compose Tooltip explicitly at each call site instead of implicitly converting every title prop into a
  tooltip?

  I have a few concerns with the current approach:

  1. Disabled buttons cannot reliably trigger the tooltip.
     Several buttons with a title can also be disabled, for example:
      - web/app/src/components/business/ConversationPane/ConversationComposer.tsx
      - web/app/src/components/business/ProfileControls/EnvKeyValueEditor.tsx

     Radix Tooltip relies on pointer and focus events from its trigger. A native disabled button cannot receive keyboard focus, and its pointer event
     behavior can also be unreliable. At minimum, these tooltips will not be accessible to keyboard users.

  2. It silently changes the meaning of the native title prop.
     ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement>, so callers reasonably expect title to be forwarded to the underlying <button>. With this
     change, title instead creates a Provider, Portal, and Tooltip without the caller explicitly requesting that behavior.

  3. Callers cannot configure the tooltip for their layout.
     The title prop only accepts a string and does not expose side, align, delayDuration, or other tooltip options. The current Tooltip defaults to the right
     side, which may not be appropriate for every button location.

  4. Each tooltip creates its own Provider.
     Creating a separate Tooltip.Provider for every button prevents tooltips from sharing delay and skip-delay behavior. A shared provider near the
     application root would provide more consistent interaction between adjacent tooltips.

  I suggest removing the Tooltip composition from Button and explicitly wrapping buttons with Tooltip at the call sites that need one. This keeps the
  primitive predictable and allows each caller to handle placement, disabled-state behavior, and accessibility appropriately.

PR #346, opened 2026-07-15
fix(web): streamline manager recreation

xxx7xxxx:
## Summary

- recreate Manager directly without the runtime and image popup
- show in-progress, success, and error feedback on the Manager page and chat action
- preserve runtime replacement behavior when recreating Manager

## Screenshots

d418c598-ee43-4e71-b723-238e031e34ff

image


web/app/src/hooks/workspace/useAgentController.ts
```diff
@@ -941,113 +933,46 @@ export function useAgentController({
...
-    setMessageActionError({ key: "", message: "" });
-    openManagerRebuildModal(managerAgent);
+    const key = `${message?.id || "message"}:${action.id}`;
+    setMessageActionBusy(key);
+    setMessageActionFeedback({ key, message: t("managerRecreateInProgress"), tone: "info" });
+    const rebuildError = await rebuildManagerFromBrowser();
+    if (rebuildError) {
+      setMessageActionFeedback({ key, message: rebuildError, tone: "error" });
+    } else {
+      setMessageActionFeedback({ key, message: t("managerRecreateSucceeded"), tone: "success" });
```
RussellLuo, line 976: `syncAgentStateUntilRunning()` resolves with `null` or a non-running agent when its timeout expires; it does not throw, and `requestManagerRebuild()` ignores that result. This branch therefore reports "Manager recreated" even when the replacement never reaches running. Could we treat a timed-out/non-running result as an error (and add a regression test) before emitting success feedback?

xxx7xxxx: Fixed. Manager recreation now treats a timeout or non-running result as an error, with regression coverage added.

RussellLuo:
LGTM

PR #350, opened 2026-07-16
fix: support hosted login and agent detail DMs

jialudev:
- Use `advertise_base_url` as the trusted OpenCSG login callback and return base for hosted sandbox deployments.
- Close the conversation agent detail drawer before navigating to the selected agent's direct message.
- Preserve local callback protections and unsaved-change confirmation behavior.


jialudev:
image

RussellLuo:
LGTM

PR #353, opened 2026-07-16
fix(models): sync OpenCSG login environment and models

jialudev:
- Sync the OpenCSG provider login with the selected environment and automatically load models from the corresponding environment after sign-in.

### DIFF
--- a/internal/api/auth.go
+++ b/internal/api/auth.go
@@ -162,8 +162,7 @@ func (h *Handler) handleAuthLogout(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if err := h.clearOpenCSGModelProviderCache(); err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
+		slog.Warn("clear OpenCSG models after logout failed", "error", err)
 	}
 	writeJSON(w, http.StatusOK, status)
 }
--- a/internal/api/auth.go
+++ b/internal/api/auth.go
@@ -5,11 +5,13 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"net"
 	"net/http"
 	"net/url"
 	"strings"
 
+	"csgclaw/internal/agent"
 	"csgclaw/internal/auth"
 )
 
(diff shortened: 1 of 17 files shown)


jialudev:
img_v3_0213l_f5bef839-05c5-440d-a58c-4bec18eaaefg

internal/api/auth.go
```diff
@@ -156,9 +161,29 @@ func (h *Handler) handleAuthLogout(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	if err := h.clearOpenCSGModelProviderCache(); err != nil {
```
RussellLuo, line 164: At this point `appAuthLogout` has already removed the auth state. If clearing or saving the model-provider cache fails, returning 500 makes the frontend treat logout as failed and keep its authenticated query state even though the server has already logged the user out. Could cache cleanup be best-effort (as in the callback refresh), or could the response still return the new logout status while surfacing the cleanup failure separately?

internal/api/model_providers.go
```diff
@@ -205,6 +208,31 @@ func (h *Handler) checkModelProvider(w http.ResponseWriter, r *http.Request) {
...
+	cfg.Models = cfg.Models.Normalized()
+	existing := cfg.Models.Providers[agent.ModelProviderIDOpenCSG].Resolved()
+	result := appCheckModelProvider(ctx, agent.ModelProviderCheckInput{
+		ID:      agent.ModelProviderIDOpenCSG,
+		BaseURL: existing.BaseURL,
+		APIKey:  existing.APIKey,
+		Headers: existing.Headers,
+		Models:  existing.Models,
+	})
+	models, changed := agent.ApplyModelProviderCheckResult(cfg.Models, agent.ModelProviderIDOpenCSG, result)
```
RussellLuo: ApplyModelProviderCheckResult only replaces Models when the check is connected and returns a non-empty list. If this callback refresh runs with cached models present and the new environment returns an empty list or the check fails, those old models remain in config; while authenticated, the frontend no longer hides them, so models from the previous environment can still be shown (and an empty successful result can even remain connected). Could this login refresh explicitly replace or clear the OpenCSG model list based on the current result?

internal/api/auth.go
jialudev: Fixed in commit 60e054af. Cache cleanup is now best-effort after appAuthLogout succeeds: cleanup failures are logged as warnings, while the handler still returns the unauthenticated status so the frontend can clear its auth and model-provider query state. Added TestHandleAuthLogoutSucceedsWhenCacheCleanupFails to cover this case.

internal/api/model_providers.go
jialudev: Fixed in commit 60e054af. The login callback refresh now clears the cached OpenCSG models, check metadata, and generated profile before applying the current environment result. This guarantees that an empty successful response or a failed check cannot retain models from the previous environment, while still persisting the new status metadata. Added regression coverage for both empty-connected and failed refresh results.

RussellLuo:
LGTM

PR #358, opened 2026-07-17
fix(agent): restrict CSGHub workers to Codex

wanghaojie124:
- When the sandbox provider is CSGHub, the template API returns only Codex worker templates.
- The agent creation UI hides the sandbox toggle and restricts runtime selection to Codex CLI.


RussellLuo:
@GatewayJ

GatewayJ:
LGTM

PR #359, opened 2026-07-17
feat(web): hide floating chat in conversations

jialudev:
- Hide the floating manager chat entry while a direct message, room, channel, or conversation is open, while keeping it available elsewhere.


jialudev:
image
image

RussellLuo:
LGTM

PR #366, opened 2026-07-17
fix(codex): prevent NFS cleanup races during recreation

xxx7xxxx:
## Summary

- Fix NFS recreation failures caused when manager bridges and a live Codex app-server kept deleted runtime files open, producing `.nfs*`, `EBUSY`, and `ENOTEMPTY`; partial cleanup could remove runtime metadata before the process was stopped.
- Stop the target bridges before deletion, stop tracked sessions even when metadata is missing, fail if the process survives termination, and prevent stale startup bridge recovery from racing the new session.
- Keep recreation synchronous and isolated to the target agent. It can still take several seconds because it waits for process shutdown, NFS removal or retries, provisioning, Codex app-server startup, handshake, and thread restoration.


xxx7xxxx:
@RussellLuo

**1. Review response**

The review correctly identified that bridges should only connect to an available session and must not implicitly restart or recreate runtimes. We removed `AgentRestarter`, and bridges now use `LiveSession` only.

However, the hydration concern no longer applies: `currentSession()` cannot hydrate or rebuild anything. Session recovery is explicitly handled during runtime startup, with failures propagated to the upper lifecycle flow.

**2. Additional improvements**

I also fix other glitches in the process of testing:

- Serialized lifecycle operations for the same agent while allowing different agents to be recreated concurrently.
- Scoped busy states and named progress messages for each agent.
- Verified the complete flow locally and in the real NFS sandbox.

RussellLuo:
LGTM

PR #367, opened 2026-07-20
feat: add real-time participant working status

jialudev:
- Add participant work leases and realtime events so conversations can report which agents are actively working.
- Show live working phases and related activity details in the conversation UI, with a compatibility fallback for older OpenClaw images.
- Integrate Codex bridge work reporting and expose participant work state through the server API.


jialudev:
image

RussellLuo:

web/app/src/hooks/workspace/useParticipantWorkStatus.ts
```diff
@@ -0,0 +1,119 @@
...
+        const agent = agents.find(
+          (candidate) =>
+            localIdentitiesMatch(candidate.id, participantID) ||
+            localIdentitiesMatch(candidate.user_id, lease.user_id) ||
+            (user ? agentMatchesUser(candidate, user) : false),
+        );
+        const id = String(user?.id || lease.user_id).trim();
+        const name = String(agent?.name || user?.name || id).trim();
+        if (id && name) {
+          byID.set(id, { id, name, requestID: lease.request_id });
```
xxx7xxxx, line 110: Not critical and not a blocker, but lease-backed status only carries `requestID` here. Non-tool activity deliveries do not always have matching `codex/openclaw.request_id` metadata, so permission/question activity from the same turn can appear in the timeline while the badge stays at generic `thinking` until a tool or final message arrives. Could we either attach request metadata to those activity deliveries or include a start-time fallback here?

RussellLuo (commented):
Overall, the separation looks solid. The API layer stays relatively lightweight and focuses on authentication, validation, error mapping, and SSE forwarding, while lease state, TTLs, revisions, epochs, tombstones, and concurrency rules remain in the dedicated worklease package. Keeping this state separate from agent busy state, task state, and IM persistence also makes the semantics clear.

One architectural boundary worth noting is that runtime integration currently differs by implementation: Codex reports leases in-process from codexbridge, while OpenClaw reports through the HTTP API. Codex uses app-server events to detect completion and derive detailed activity, but the lease itself is not a shared runtime or app-server capability. As a result, a Codex turn invoked through another entry point without codexbridge would not report participant work status.

This seems reasonable for the current scope because the bridge owns the room, participant, and source-message metadata. If the feature later expands to more runtimes, entry points, or lifecycle states, a shared runtime turn-lifecycle/reporting abstraction could centralize the start, renew, completion, failure, and cancellation semantics while keeping the current in-process and HTTP paths as adapters. This is not a blocker, but it may be useful to document as an intentional boundary or follow-up.

internal/api/participant_work.go
```diff
@@ -0,0 +1,155 @@
...
+	var extra any
+	if err := decoder.Decode(&extra); errors.Is(err, io.EOF) {
+		return nil
+	} else if err != nil {
+		return fmt.Errorf("decode request: %w", err)
+	}
+	return fmt.Errorf("decode request: multiple JSON values")
+}
+
+func validUUID(value string) bool {
```
RussellLuo: Could we avoid maintaining a custom UUID parser here? `github.com/google/uuid` is already present in `go.mod`, and the lease ID format belongs to the `worklease` domain. I suggest keeping both generation and validation behind that package:

```go
// internal/worklease
func NewID() string {
    return uuid.NewString()
}

func ValidID(value string) bool {
    if len(value) != 36 {
        return false
    }
    _, err := uuid.Parse(value)
    return err == nil
}
```

The length check preserves the current requirement for the canonical 36-character form, since `uuid.Parse` also accepts some alternative representations. The API validation can then be reduced to `participantID == "" || !worklease.ValidID(leaseID)`, keeping the Handler focused on HTTP concerns and making `worklease` the single owner of lease ID generation and validation. Importing the package directly would also promote the existing `google/uuid` entry from indirect to direct dependency.

jialudev: Fixed in 9dd6c29a. Lease ID generation and canonical validation now live in `internal/worklease` via `github.com/google/uuid`, and the API delegates validation to `worklease.ValidID`. I also added focused tests for canonical and alternative UUID forms.

web/app/src/hooks/workspace/useParticipantWorkStatus.ts
jialudev: Thanks for calling this out. In the current UX, when Codex is waiting for a permission or question response, the UI presents the corresponding interaction card for the user to act on; that state is not intended to describe what the agent is doing through the badge. The generic `thinking` badge during this interval therefore does not affect the current flow. Since this is non-blocking, we would prefer to defer adding request metadata or a start-time fallback until we have a concrete Codex requirement, to avoid introducing extra correlation logic and potential regressions in this change.

xxx7xxxx:
LGTM

web/app/src/hooks/workspace/useParticipantWorkStatus.ts
xxx7xxxx: Got it

RussellLuo:
LGTM

PR #368, opened 2026-07-20
feat(connectors): add GitLab PAT integration

wanghaojie124:


RussellLuo (commented):
The embedded GitLab skill is quite large and includes several internally customized workflows. Could we move the entire skill to CSGHub SkillHub and provide it as an installable remote skill instead of embedding it directly in CSGClaw? This would keep the CSGClaw core distribution smaller, avoid shipping organization-specific workflows to every user, and allow the skill to evolve independently of CSGClaw releases.

internal/template/embed/manager/codex/workspace/skills/gitlab/scripts/get_project.py
```diff
@@ -0,0 +1,90 @@
...
+Usage: python3 get_project.py "group/project"
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from urllib.parse import quote
+
+import httpx
```
xxx7xxxx: This skill is shipped as embedded manager code, but many of the new GitLab helper scripts depend on `httpx` while the instructions invoke them with bare `python3` and no dependency bootstrap. I can reproduce a failure before argument parsing with `python3 .../get_project.py --help` when the user-site `httpx`/`click` install is incompatible, so the first GitLab workflow can fail before it ever requests a connector lease. Please either make these helpers stdlib-only, vendor/pin dependencies, or run them through a controlled environment.

internal/template/embed/manager/codex/workspace/skills/gitlab/scripts/gitlab_utils.py
```diff
@@ -0,0 +1,221 @@
...
+ENV_CSGCLAW_ACCESS_TOKEN = "CSGCLAW_ACCESS_TOKEN"
+MANAGER_GITLAB_CREDENTIAL_PATH = "/api/v1/agents/agent-manager/connectors/gitlab/credential"
+
+_CREDENTIAL_MISSING_MSG = (
+    "CSGClaw GitLab Connector 未配置或无法获取 lease。请用户在 Connector 中检查 Base URL 与 Token；"
+    "不要要求用户在对话中发送 Token。"
+)
+DEFAULT_GIT_AUTH_USERNAME = "oauth2"
+
+_REDACT_PATTERNS = (
```
xxx7xxxx: `run_with_gitlab_auth.py` prints child stdout/stderr after calling `_redact_secret`, but this redactor only covers a few header-shaped strings. It does not redact the actual lease token or common env output such as `GITLAB_TOKEN=...`, so a failing child command that dumps env or echoes the token can leak the connector PAT into logs. Please redact exact lease token matches and common env/header spellings before printing subprocess output.

xxx7xxxx:
Agree with @RussellLuo . This GitLab skill is large, organization-specific, and would be better shipped as a CSGHub SkillHub remote skill instead of being embedded in CSGClaw core. GitHub PR Case is different since it aims to reach a best practice workflow.

internal/template/embed/manager/codex/workspace/skills/gitlab/scripts/get_project.py
xxx7xxxx: And should we consider how to handle third-party module installation for Skill? Such as httpx in this GitLab Skill. Or we just ignore it and hope the LLM handles it itself.

wanghaojie124: Thanks for catching this. Resolved. The GitLab skill has been removed from the embedded Manager template and moved to the stg SkillHub as an independently installable remote skill. We also removed the `httpx` dependency from the remote skill helpers and replaced it with a standard-library HTTP implementation. The scripts were verified with `python3 -S ... --help`, so they no longer depend on user-site packages or LLM-driven dependency installation.

internal/template/embed/manager/codex/workspace/skills/gitlab/scripts/gitlab_utils.py
wanghaojie124: Thanks for pointing this out. Resolved in the remote GitLab skill. The wrapper now redacts exact lease-token matches, URL-encoded token values, and common environment/header forms such as `GITLAB_TOKEN=...`, `PRIVATE-TOKEN: ...`, and `Authorization: Bearer ...` before forwarding child stdout or stderr. It also removes `CSGCLAW_ACCESS_TOKEN` from the child environment after obtaining the lease.

wanghaojie124:
Thanks @RussellLuo and @xxx7xxxx. This has been addressed in `a4a2c552`. The organization-specific GitLab workflow has been removed from the CSGClaw Manager template and published separately in the stg SkillHub. CSGClaw now retains only the platform-level GitLab Connector and credential-lease security rules. The generated Manager instructions no longer hard-code `skills/gitlab/SKILL.md`, and the remote skill can evolve independently of CSGClaw releases. Tests were updated to cover the Connector guidance and remote-path basename mapping without depending on a live SkillHub environment.

internal/agentmanager/connectors.go
```diff
@@ -52,7 +53,8 @@ type DefaultConnectorGrantPolicy struct{}
 func (DefaultConnectorGrantPolicy) AllowsConnectorCredential(_ context.Context, ref AgentConnectorRef, provider string) bool {
 	return strings.TrimSpace(ref.AgentID) == agent.ManagerUserID &&
 		strings.EqualFold(strings.TrimSpace(ref.AgentRole), agent.RoleManager) &&
-		strings.EqualFold(strings.TrimSpace(provider), connectors.ProviderGitHub)
+		(strings.EqualFold(strings.TrimSpace(provider), connectors.ProviderGitHub) ||
+			strings.EqualFold(strings.TrimSpace(provider), connectors.ProviderGitLab))
```
xxx7xxxx, line 57: [P1] Adding GitLab here exposes the PAT through the existing agent credential route without authenticating the caller as the manager. The API handler only validates the shared server bearer token and then trusts the `{id}` path segment; Codex runtimes receive `CSGCLAW_ACCESS_TOKEN`, so a worker can call `/api/v1/agents/agent-manager/connectors/gitlab/credential` and satisfy this policy as the manager. Please bind the lease request to the authenticated agent identity, or issue per-agent capability tokens, before enabling GitLab here.

internal/runtime/codex/runtime_test.go
```diff
@@ -1657,7 +1657,7 @@ func TestRuntimeCreateInstallsManagerTemplate(t *testing.T) {
 	}
 
 	skillsRoot := filepath.Join(root, agent.ManagerUserID, ".codex", "home", "skills")
-	for _, name := range []string{"agent-creator", "agent-teams", "feishu"} {
+	for _, name := range []string{"agent-creator", "agent-teams", "feishu", "gitlab"} {
```
xxx7xxxx: [P1] This expectation still requires an embedded `skills/gitlab/SKILL.md`, but the PR now removes the embedded GitLab skill and moves it to SkillHub. I ran `go test ./internal/runtime/codex` on `a4a2c552` and it fails here with `manager template skill "gitlab" missing`. Please drop `gitlab` from the embedded manager-template assertions, or reintroduce a matching embedded skill if the template is still supposed to ship it.

internal/agentmanager/connectors.go
xxx7xxxx: @wanghaojie124 You can decide whether to fix this issue or not. IMO, it's not that critical since other agents could also get the credentials.

wanghaojie124: Fixed in d0e3aa7f. The credential route now requires a Manager-only connector capability in addition to the shared server bearer token. The capability is generated from a per-service random key, injected only into the Manager runtime as CSGCLAW_CONNECTOR_CAPABILITY, and validated against the requested agent identity using a constant-time comparison. Workers do not receive this capability, so a Worker cannot obtain a GitHub or GitLab lease by reusing CSGCLAW_ACCESS_TOKEN and spoofing the agent-manager path. I also added tests covering Manager access, Worker impersonation, and missing-capability rejection.

internal/runtime/codex/runtime_test.go
wanghaojie124: Fixed in d0e3aa7f. I removed gitlab from the embedded Manager skill assertions because the GitLab skill is now distributed through SkillHub instead of being shipped in the Manager template. go test ./internal/runtime/codex passes.

internal/agentmanager/connectors.go
xxx7xxxx: Can workers running in codex access GitLab via the connectors? I think it should allow them to access it since connector status is in global scope.

wanghaojie124: > Can workers running in codex access GitLab via the connectors? I think it should allow them to access it since connector status is in global scope.

Yes, just install gitlab skill

xxx7xxxx (approved):

PR #371, opened 2026-07-21
fix(web): clear agent profile errors on agent switch

youngbeom-shin:
## Summary

- Clear `agentMCPAddError` and `agentMCPDeleteError` when switching agents (via `agentDetailAgentID` useEffect), matching existing behavior for skill add/delete errors
- Clear `agentPageError` (saveError) when switching agents even when there are unsaved changes, preventing stale error messages from persisting on the new agent's profile page


jialudev:
LGTM

PR #373, opened 2026-07-21
fix(feishu): avoid worker registration overwriting admin

GatewayJ:
## Summary
- prevent worker Feishu registration finalize from creating or overwriting the global feishu:admin participant
- keep manager registration behavior unchanged, including admin OpenID/name binding
- update finalize tests for worker/admin role boundaries


xxx7xxxx (approved):
LGTM

PR #377, opened 2026-07-21
fix(agent): default avatar for agents without IM user avatar

youngbeom-shin:
## Summary

- When an agent has no IM user avatar (e.g., manager-created workers), resolveAgentAvatarSource now returns the default agent avatar (avatar/3D-1.png) instead of an empty string, so the agent always displays a proper built-in avatar image rather than a plain text fallback.


youngbeom-shin:
截屏2026-07-22 18 16 13
agents with random avatar that created by manager

jialudev (approved):
LGTM

PR #379, opened 2026-07-21
refactor(template): reorganize agent template layout

wanghaojie124:
image
image
image
image


internal/config/config.go
```diff
@@ -335,7 +335,7 @@ const (
 	DefaultOfficialHubRegistryName  = "official"
 	DefaultOfficialHubRegistryURL   = "https://hub.opencsg.com"
 	DefaultBootstrapManagerTemplate = "builtin.manager-codex"
-	DefaultBootstrapWorkerTemplate  = "builtin.picoclaw-worker"
+	DefaultBootstrapWorkerTemplate  = "builtin.openclaw-worker"
```
RussellLuo, line 338: This default change leaves the repository's Go suite red: `go test ./...` now fails `TestFormatEffectiveConfigPrintsExpandedMaskedEnvValues`, `TestLoadConfigRewritesLegacyBootstrapTemplateRefs`, and `TestFormatEffectiveConfigFormatsSectionsWithoutExtraWhitespace`, all of which still expect `builtin.picoclaw-worker`. Please update the cross-package expectations (and validate the intended migration) before merging.

internal/template/local_store.go
```diff
@@ -99,18 +105,24 @@ func (s *LocalStore) FetchWorkspace(_ context.Context, id string) (WorkspaceRef,
...
 		return WorkspaceRef{}, err
 	}
-	workspace := s.workspaceRoot(id)
-	info, err := os.Stat(workspace)
-	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			return WorkspaceRef{}, nil
-		}
-		return WorkspaceRef{}, fmt.Errorf("stat local hub workspace %q: %w", workspace, err)
+	if _, err := os.Stat(filepath.Join(s.templateRoot(id), localInstructionsDirName)); errors.Is(err, os.ErrNotExist) {
```
RussellLuo: Existing local templates were persisted as `<template>/workspace/...`. After an upgrade, this new `instructions/` check returns an empty workspace ref for those templates: the manifest still appears in Hub, but creating from it silently drops the saved workspace. Please add a legacy fallback/migration and a regression test before switching the on-disk format.

internal/template/local_store.go
```diff
@@ -99,18 +105,24 @@ func (s *LocalStore) FetchWorkspace(_ context.Context, id string) (WorkspaceRef,
...
+		return WorkspaceRef{}, nil
 	}
-	if !info.IsDir() {
-		return WorkspaceRef{}, fmt.Errorf("local hub workspace %q is not a directory", workspace)
+	return materializeTemplateDir(s.templateRoot(id))
```
RussellLuo: `materializeTemplateDir` creates a fresh `csgclaw-template-workspace-*` directory, but `templateWorkspaceCleanup` still only removes builtin and remote refs and explicitly skips local ones. Every agent created from a local template therefore leaves a full materialized workspace in the temp directory. Please make ownership explicit and clean this ref up (or keep returning a persistent local path).

web/app/src/pages/HubPage/components/HubDetailPane/HubDetailPane.tsx
```diff
@@ -463,6 +602,67 @@ export function HubDetailPane({
...
+    () => templateSkillSummaries(workspaceEntries, workspaceFile, workspaceFiles),
+    [workspaceEntries, workspaceFile, workspaceFiles],
+  );
+  const templateMCPServers = useMemo(
+    () => templateMCPServerSummaries(workspaceFile, workspaceFiles),
+    [workspaceFile, workspaceFiles],
+  );
+  const templateInstructions =
+    typeof selectedTemplate === "object" && selectedTemplate
+      ? String((selectedTemplate as { instructions?: string | null }).instructions || "")
```
RussellLuo: The Hub template API/type does not expose an `instructions` field, so this assertion always produces an empty string. `selectTemplateTab` fetches `instructions/AGENTS.md`, but the textarea never reads `workspaceFile`, which makes the new Instructions tab blank for every template. Please derive the value from the loaded workspace file and assert its content in the component test.

internal/template/template_layout.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	dstRoot, err := mkdirHubWorkspaceTemp("csgclaw-template-workspace-*")
+	if err != nil {
+		return WorkspaceRef{}, err
+	}
+	cleanup := func(err error) (WorkspaceRef, error) {
+		_ = os.RemoveAll(dstRoot)
+		return WorkspaceRef{}, err
+	}
+	if err := copyWorkspaceTreeFS(srcFS, instructionsRoot, dstRoot, "template instructions"); err != nil {
```
RussellLuo, line 50: Template instructions are being materialized directly into the workspace, so for Codex `instructions/AGENTS.md` becomes `.codex/workspace/AGENTS.md` and is treated as project-scoped guidance. The template contract should be agent-global instead: seed the template content as the base of the agent's `CODEX_HOME/AGENTS.md`, do not pass request/profile instructions during template-based creation, and only add or replace the managed profile-instructions block in that same file on later profile edits. The template `AGENTS.md` should not be installed at workspace scope. Please separate template instructions from `WorkspaceOverlay` and cover creation plus profile-update behavior with regression tests.

RussellLuo:
It would be useful for both the template and Agent Profile Instructions pages to support two views of the same underlying `AGENTS.md`.

**Default mode**

- Show only the user-facing custom instructions.
- Save them to the managed instructions block.

**Advanced mode**

- Expand and edit the complete effective `AGENTS.md`, including the template/base content and managed profile block.
- Preserve lossless round-tripping between the two modes.

The document model should remain layered:

```text
template/base AGENTS.md
+ profile instructions managed block
= effective AGENTS.md
```

On the template page, advanced edits should update `instructions/AGENTS.md`. On the Agent Profile page, advanced edits should update the effective runtime `AGENTS.md`, while the default mode continues to update only the managed instructions block.

The runtime destinations would be `CODEX_HOME/AGENTS.md` for Codex and the workspace `AGENTS.md` for OpenClaw.

This keeps the common customization flow simple while still allowing advanced users to precisely control the full `AGENTS.md`.

RussellLuo:
Two flows are closely related to the template contract introduced by this PR and should be considered together:

1. **Recreating an agent** should preserve the agent's current runtime-managed instructions, skills, and MCP configuration instead of falling back to only the original template state.
2. **Publishing an agent as a template** should export those runtime assets into the canonical template layout.

The export needs to be runtime-aware. For example:

- Codex should read the effective `AGENTS.md`, skills, and MCP configuration from the agent's `CODEX_HOME` and its runtime configuration.
- OpenClaw should read `AGENTS.md` and skills from its workspace, plus MCP servers from its runtime configuration.

Both should then be normalized into the same runtime-independent template structure:

```text
instructions/AGENTS.md
skills/...
mcps/mcp.json
```

The conversion should also define how generated/managed instruction blocks are normalized and ensure that credentials, tokens, and other runtime-only state are not published into the template.

Covering these round-trip paths will make templates reliable not only for initial creation, but also for recreating customized agents and deriving new templates from running agents.

internal/config/config.go
wanghaojie124: Addressed in 4a682de3. I updated the cross-package config expectations from `builtin.picoclaw-worker` to `builtin.openclaw-worker`, including effective-config formatting and legacy bootstrap rewrite coverage. The three targeted `cli/serve` tests now pass.

internal/template/local_store.go
wanghaojie124: Addressed in 4a682de3. Local templates now detect both the canonical layout and the legacy `<template>/workspace/...` layout. Legacy workspaces are materialized through a compatibility path, with regression coverage verifying that the saved `AGENTS.md` and skills remain available after upgrade.

internal/template/local_store.go
wanghaojie124: Addressed in 4a682de3. `WorkspaceRef` now carries explicit temporary ownership, and cleanup removes every caller-owned materialized workspace regardless of registry kind. Added regression coverage for local-template cleanup and successful materialization cleanup.

web/app/src/pages/HubPage/components/HubDetailPane/HubDetailPane.tsx
wanghaojie124: Addressed in 4a682de3. The Instructions pane now reads `instructions/AGENTS.md` from the loaded workspace file/cache and proactively loads that file for the selected template. The component test asserts the actual file content and covers switching between the default managed-instructions view and the complete advanced document.

internal/template/template_layout.go
wanghaojie124: Addressed in 4a682de3. Codex template instructions are separated from `WorkspaceOverlay` and seeded as the base of `CODEX_HOME/AGENTS.md`; request/profile instructions are suppressed during template creation. Later profile edits add or replace only the managed block in that same file, while the template base remains intact and is not copied into the workspace. Added service and Codex runtime regression tests for creation and subsequent profile update behavior.

wanghaojie124:
Addressed in 4a682de3. Following up on the layered Instructions proposal in https://github.com/OpenCSGs/csgclaw/pull/379#issuecomment-5034716945:

- Both the template and Agent Profile Instructions pages now provide Default and Advanced views.
- Default mode edits only the user-facing instructions in the managed block.
- Advanced mode edits the complete effective `AGENTS.md` and preserves the template/base content during round trips.
- Template advanced edits persist to `instructions/AGENTS.md`; Agent Profile advanced edits persist to the runtime document.
- Runtime destinations are now explicit: `CODEX_HOME/AGENTS.md` for Codex and workspace `AGENTS.md` for OpenClaw.

The UI component tests and TypeScript checks pass.

wanghaojie124:
Addressed in 4a682de3. Following up on the recreate/publish round-trip proposal in https://github.com/OpenCSGs/csgclaw/pull/379#issuecomment-5034742988:

- Recreate preserves the current effective instructions, custom skills, and managed MCP configuration.
- Publish is runtime-aware: Codex exports instructions and skills from `CODEX_HOME`, while OpenClaw exports them from its workspace.
- Both normalize into `instructions/AGENTS.md`, `skills/...`, and `mcps/mcp.json`.
- Runtime-only MCP fields such as environment, headers, tokens, secrets, passwords, and API keys are removed before publishing.
- Regression coverage was added for runtime-aware export, local materialization ownership, and Codex layered instruction behavior.

Targeted template, agent, Codex, and OpenClaw tests pass.

RussellLuo:
LGTM

PR #382, opened 2026-07-22
fix(packaging): bundle and install companion csgclaw-cli

wanghaojie124:


GatewayJ:
1. CSGClaw 基础镜像基础镜像是否也需要准备host版本cli
2. sandbox-runtime dockerfile 需要在构造 bundle seed 时同时写入 csgclaw_dir/host/csgclaw-cli,agentihub的安装流程也需要对应修改

RussellLuo (commented):
Could we simplify this by treating the host `csgclaw-cli` as a normal companion binary and reserving `sandbox-tools` only for the CLI mounted into sandboxes?

Each release archive is already specific to one host OS and architecture, so the host binary does not need another `<os>-<arch>` directory. A bundle could use this layout:

```text
csgclaw/
├── .csgclaw-bundle.json
└── bin/
    ├── csgclaw[.exe]
    ├── csgclaw-cli[.exe]          # native host CLI
    ├── boxlite[.exe]               # when supported
    └── sandbox-tools/
        └── csgclaw-cli             # Linux CLI mounted into the sandbox
```

The installer could then expose both host commands through the existing `INSTALL_DIR`, using the same PATH behavior for each:

```text
bundle/bin/csgclaw[.exe]
  -> INSTALL_DIR/csgclaw[.exe]

bundle/bin/csgclaw-cli[.exe]
  -> INSTALL_DIR/csgclaw-cli[.exe]

bundle/bin/sandbox-tools/csgclaw-cli
  -> ~/.csgclaw/sandbox-tools/csgclaw-cli
```

This keeps the responsibilities clear: the host CLI is installed and upgraded together with `csgclaw`, while `~/.csgclaw/sandbox-tools` remains an internal runtime directory used only by the existing sandbox mount logic. Codex can simply inherit the normal PATH, so `HostToolsDir`, `DefaultHostToolsDir`, the host-specific refresh fields and synchronization logic, and the runtime-specific PATH injection can all go away.

During an upgrade, replacing the bundle updates both host commands together, and only the sandbox CLI needs to be synchronized into `~/.csgclaw/sandbox-tools`. For compatibility with older bundles, the sandbox asset resolver can temporarily fall back from the new path:

```text
bin/sandbox-tools/csgclaw-cli
```

to the existing path:

```text
bin/csgclaw_dir/csgclaw-cli
```

The standalone `csgclaw-cli_<version>_<os>_<arch>` archive can still be retained for users who only need the CLI, while the default `csgclaw` installer installs both commands.

wanghaojie124:
@RussellLuo Thanks, this direction makes sense and I’ve adjusted the implementation to follow it.

Current layout is now:

```text
csgclaw/bin/
  csgclaw[.exe]
  csgclaw-cli[.exe]        # native host companion
  boxlite[.exe]
  sandbox-tools/
    csgclaw-cli            # Linux CLI mounted into sandboxes

RussellLuo:
LGTM

PR #388, opened 2026-07-22
fix(web): localize remote skill replacement action

GatewayJ:
## Summary

- Add Chinese and English labels for replacing an installed remote skill.
- Cover the replacement action label with an i18n regression test.

## Validation

- pnpm --dir web/app exec vitest run tests/shared/i18n.test.ts
- pnpm --dir web/app typecheck
- pnpm --dir web/app lint
- pnpm --dir web/app format:check


jialudev:
LGTM

PR #395, opened 2026-07-23
fix(web): scope MCP creation errors to dialog

youngbeom-shin:
## Summary
- keep MCP creation errors scoped to the add dialog
- preserve update and delete errors in their existing detail contexts
- clear stale creation errors when the add dialog opens or closes
- add a regression test that asserts the duplicate-name error renders once


youngbeom-shin:
截屏2026-07-23 17 10 00

jialudev (approved):
LGTM

PR #400, opened 2026-07-24
fix(web): add stop control to floating chat

jialudev:
- Expose the existing turn-stop action in the floating manager chat so active replies can be stopped without opening the full conversation view.
- Cover the floating composer stop interaction with a regression test.


jialudev:
image

PR #406, opened 2026-07-27
feat(mcp): add remote MCP server installation

GatewayJ:
## Summary

- add authenticated remote MCP discovery and server-side installation from the effective official Hub
- add Remote install UI, pagination/search support, and MCP configuration normalization shared by supported runtimes
- document the remote MCP API and include API, Go, and frontend coverage

## Screenshot

Remote MCP installation dialog, captured with Playwright using demonstration data and the MCP resource page behind the modal:

Remote MCP installation dialog

## Validation

- `go test ./internal/api ./internal/mcp ./internal/mcpschema ./internal/runtime/...`
- `pnpm --dir web/app exec vitest run tests/components/HubDetailPane.test.tsx tests/api/mcp.test.ts tests/models/mcp.test.ts --maxWorkers=1` *(14 passed; 2 pre-existing template-instructions assertions fail because their fixture has no `source.kind: "local"`)*

## Notes

- The browser receives only remote MCP summaries; full configuration and headers are fetched server-side during installation.


RussellLuo:
@wanghaojie124 Please help review this.

internal/mcp/remote.go
```diff
@@ -0,0 +1,372 @@
...
+		return ""
+	}
+}
+
+func remoteServerID(value any) string {
+	switch typed := value.(type) {
+	case string:
+		return strings.TrimSpace(typed)
+	case float64:
+		return strconv.FormatInt(int64(typed), 10)
```
wanghaojie124: Numeric IDs are decoded through `any` as `float64` and then cast to `int64`. IDs above 2^53 lose precision (and fractional values are silently truncated), so the subsequent detail request can target the wrong ID or return 404. Please preserve the JSON number exactly, for example with `json.Decoder.UseNumber`, and reject non-integer/out-of-range values.

internal/api/mcp_remote.go
```diff
@@ -0,0 +1,147 @@
...
+			Description: item.Description,
+			ID:          item.ID,
+			Name:        item.Name,
+			Protocol:    item.Protocol,
+			URL:         item.URL,
+		})
+	}
+	writeJSON(w, http.StatusOK, remoteMCPServersListResponse{
+		Items:    items,
+		NextPage: nextRemoteMCPServersPage(page, per, list.Total, len(items)),
```
wanghaojie124: When the Hub omits `total`, this uses the number of filtered summaries to decide whether another page exists. `decodeRemoteServerPage` silently drops records with a missing/invalid ID or name, so one malformed record in a full upstream page makes `len(items) < per` and prematurely ends pagination, hiding all later servers. Please base this decision on the raw record count (or an upstream next-page signal), not the filtered response length.

wanghaojie124:
LGTM

PR #407, opened 2026-07-27
feat(desktop): add Electron application runtime

jialudev:
- Add a secure Electron desktop shell that reuses the existing Go service and Web UI, including sidecar lifecycle, OAuth, navigation, and desktop-managed updates.


jialudev:
image

cli/serve/desktop.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	if err := ensureServeBootstrapState(ctx, configPath, false); err != nil {
+		return err
+	}
+	cfg, err := loadConfig(configPath)
+	if err != nil {
+		return err
+	}
+
+	listener, err := net.Listen("tcp", "127.0.0.1:0")
```
RussellLuo: This loopback-only listener is not reachable at the sandbox URL generated on Linux (the default Docker provider falls back to a host LAN IP there), nor by non-Docker providers. Allowing that value in `ServerAccessHosts` only validates the Host header; it does not make the socket listen on that interface, so desktop agents cannot call the manager API. Please provide a sandbox-reachable listener/proxy while keeping the Renderer endpoint loopback-only, and cover Linux/non-Docker desktop connectivity.

desktop/src/main/appLifecycle.ts
```diff
@@ -0,0 +1,261 @@
...
+    }
+    if (confirm && !(await this.confirmQuit())) {
+      return;
+    }
+    this.quitting = true;
+    this.cleanup();
+    this.windowManager?.destroy();
+    this.tray?.destroy();
+    this.tray = null;
+    await this.supervisor?.stop("app-quit");
```
xxx7xxxx, line 92: `stop()` only shuts down the Go sidecar. `agent.Service.Close()` closes runtime handles, while Docker and BoxLite `Runtime.Close()` are no-ops, so sandbox agents remain running despite the quit confirmation. The next launch binds a new ephemeral port, and `StartConfiguredAgents` skips running sandbox agents, leaving them configured against the dead old port. Please explicitly stop these agents on desktop quit, or reconcile and restart them before reporting the new sidecar ready.

desktop/src/main/windowManager.ts
```diff
@@ -0,0 +1,136 @@
...
+  private mainWindow: BrowserWindow | null = null;
+  private readonly desktopSession: Session;
+
+  constructor(private readonly options: WindowManagerOptions) {
+    this.desktopSession = session.fromPartition("persist:csgclaw-desktop", { cache: true });
+    this.installRequestAuthentication();
+  }
+
+  setConnection(baseURL: string, sessionToken: string): void {
+    this.allowedOrigin = new URL(baseURL).origin;
```
xxx7xxxx, line 23: `baseURL` contains a freshly allocated sidecar port, so the renderer gets a different origin on every app launch. Web Storage is origin-scoped including the port; the persistent partition therefore cannot recover the theme, locale, auth environment, Feishu registration records, layout preferences, and other state stored in `localStorage`. Please provide a stable renderer origin or move and migrate this state through a stable desktop-owned store.

web/app/src/shared/platform/desktopBridge.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+export type DesktopUpdateStatus = {
+  state: "idle" | "checking" | "available" | "not-available" | "downloaded" | "error" | "unsupported";
+  currentVersion: string;
+  availableVersion?: string;
+  message?: string;
+};
+
+export type DesktopBridge = {
```
xxx7xxxx, line 18: (nit): The desktop bridge has no sidecar restart operation, so the existing settings flow still calls `/api/v1/server/restart`. `_desktop-serve` has no daemon PID file, causing the restart helper to report `manual_restart_required` and direct desktop users to stop and rerun the CLI; changed settings are not applied by the managed app. Please add a supervised sidecar restart path and route the config controller through it in desktop mode.

desktop/forge.config.ts
```diff
@@ -0,0 +1,175 @@
...
+
+const config: ForgeConfig = {
+  packagerConfig: {
+    appBundleId: "com.opencsg.csgclaw.desktop",
+    appCategoryType: "public.app-category.developer-tools",
+    appVersion: desktopVersion,
+    asar: true,
+    executableName: "CSGClaw",
+    extraResource: [backendResources, updateConfig],
+    icon: appIcon,
```
xxx7xxxx, line 67: (nit): `appIcon` is always the macOS `.icns` file. Electron Packager normalizes Windows icons to `.ico`, which is absent here, while `MakerDeb` receives no Linux icon and uses its generic default. Please add and wire platform-specific `.icns`, `.ico`, and `.png` assets so Windows and Linux installers ship with the CSGClaw icon.

cli/serve/desktop.go
jialudev: > This loopback-only listener is not reachable at the sandbox URL generated on Linux (the default Docker provider falls back to a host LAN IP there), nor by non-Docker providers. Allowing that value in `ServerAccessHosts` only validates the Host header; it does not make the socket listen on that interface, so desktop agents cannot call the manager API. Please provide a sandbox-reachable listener/proxy while keeping the Renderer endpoint loopback-only, and cover Linux/non-Docker desktop connectivity.

Added a separate sandbox listener while keeping the Renderer listener loopback-only.

Added tests for Linux Docker and non-Docker providers.

Fixed in 79243041.

desktop/src/main/appLifecycle.ts
jialudev: > `stop()` only shuts down the Go sidecar. `agent.Service.Close()` closes runtime handles, while Docker and BoxLite `Runtime.Close()` are no-ops, so sandbox agents remain running despite the quit confirmation. The next launch binds a new ephemeral port, and `StartConfiguredAgents` skips running sandbox agents, leaving them configured against the dead old port. Please explicitly stop these agents on desktop quit, or reconcile and restart them before reporting the new sidecar ready.

Running sandbox agents are now stopped before the desktop sidecar shuts down or restarts.  Fixed in 46713ebf

desktop/src/main/windowManager.ts
jialudev: > `baseURL` contains a freshly allocated sidecar port, so the renderer gets a different origin on every app launch. Web Storage is origin-scoped including the port; the persistent partition therefore cannot recover the theme, locale, auth environment, Feishu registration records, layout preferences, and other state stored in `localStorage`. Please provide a stable renderer origin or move and migrate this state through a stable desktop-owned store.

The Renderer listener now uses the stable loopback address 127.0.0.1:18791. Fixed in 46713ebf.

web/app/src/shared/platform/desktopBridge.ts
jialudev: > (nit): The desktop bridge has no sidecar restart operation, so the existing settings flow still calls `/api/v1/server/restart`. `_desktop-serve` has no daemon PID file, causing the restart helper to report `manual_restart_required` and direct desktop users to stop and rerun the CLI; changed settings are not applied by the managed app. Please add a supervised sidecar restart path and route the config controller through it in desktop mode.

Added a supervised sidecar restart through Electron IPC. Desktop settings no longer use the daemon restart API. Fixed in 46713ebf.

desktop/forge.config.ts
jialudev: > (nit): `appIcon` is always the macOS `.icns` file. Electron Packager normalizes Windows icons to `.ico`, which is absent here, while `MakerDeb` receives no Linux icon and uses its generic default. Please add and wire platform-specific `.icns`, `.ico`, and `.png` assets so Windows and Linux installers ship with the CSGClaw icon.

Added platform-specific .icns, .ico, and .png icons. Fixed in 46713ebf.

xxx7xxxx (approved):

RussellLuo:
LGTM

PR #413, opened 2026-07-29
feat(chat): add room-wide agent notifications

xxx7xxxx:
## Summary

- Add a room-level notify-all switch that fans human messages out to every agent while preventing agent reply loops. Default value is false.
- Preserve participant working status and persisted room state through isolated startup tests.
- Harden Codex app-server shutdown against concurrent request writes.
- This latest-message-wins behavior remains unsolved for now.

## Screenshots

image

image


RussellLuo:
@wanghaojie124 @jialudev Please help review this.

internal/api/conversation_command.go
```diff
@@ -89,13 +90,14 @@ func newConversationTargets(room im.Room, message im.Message, isAgent func(strin
 	if isAgent == nil {
 		return nil
 	}
+	notifyAllAgents := room.NotifyAllAgents && !isAgent(message.SenderID)
```
wanghaojie124, line 93: Could we confirm the intended behavior here? The API docs and UI copy say notify-all wakes every agent when a human message has no explicit mentions, but this condition enables fan-out for every human message. As a result, a message such as `@agent-a please handle this` also wakes agents B and C. If explicit mentions should take precedence, please gate fan-out on the absence of mentions (including supported `<at>` tags) and add a regression test. Otherwise, the docs and UI copy should say that all human messages are broadcast.

xxx7xxxx: Confirmed: notify-all intentionally broadcasts every human message, including messages with explicit mentions. Updated the API docs and UI copy and added regression coverage.

wanghaojie124:
LGTM

web/app/src/models/conversations.ts
```diff
@@ -873,6 +874,7 @@ export function applyIMEvent<T extends IMData | null | undefined>(
     (event.type === "conversation.created" ||
       event.type === "conversation.members_added" ||
       event.type === "room.created" ||
+      event.type === "room.updated" ||
```
RussellLuo, line 877: `room.updated` is applied by replacing the whole conversation, but `UpdateRoom` snapshots the room before unlocking and publishes afterward. A concurrent `message.created` can therefore be applied first and then overwritten by this older room snapshot, making the new message disappear from the UI (the PATCH response has the same replacement path). Please merge only `notify_all_agents` for this mutation, or add versioned/ID-based reconciliation, and cover the out-of-order event case.

internal/runtime/codex/appserver_client.go
```diff
@@ -281,10 +293,16 @@ func (c *appServerClient) closeAllPending(err error) {
 	if err == nil {
 		err = fmt.Errorf("codex app-server client closed")
 	}
+	c.writeMu.Lock()
```
RussellLuo, line 296: `closeAllPending` now waits for `writeMu`, while `writeJSONLine` holds that mutex across `stdin.Write`. If the app-server is alive but no longer reading stdin and the pipe write blocks, `Stop` gets stuck here before it can close stdin, signal the process, or enter its timeout select. Please close or cancel the pipe before waiting for the writer, or redesign this synchronization so shutdown cannot depend on a blocked write; a blocking-writer regression test would cover the failure mode.

web/app/src/models/conversations.ts
xxx7xxxx: Fix it in https://github.com/OpenCSGs/csgclaw/pull/419

internal/runtime/codex/appserver_client.go
xxx7xxxx: Ditto

PR #418, opened 2026-07-29
feat(hub): publish agent templates to remote registry

wanghaojie124:


RussellLuo:
At the product level, we plan to support publishing to both local and remote registries. Let’s wait for the product design document tomorrow; the code may need further adjustments once the design is finalized.

PR #419, opened 2026-07-29
fix: prevent stale room updates and shutdown hangs

xxx7xxxx:
## Summary

- Preserve current room messages when notify-all updates arrive from events or PATCH responses.
- Prevent Codex app-server shutdown from waiting behind blocked stdin writes.


xxx7xxxx:
@RussellLuo Any concerns?

RussellLuo:
LGTM

PR #423, opened 2026-07-30
fix(app): Honor system proxy like other native apps & use the standard port 18080 for the Electron

xxx7xxxx:
## Summary

- Honor system proxy like other native apps
- use the standard port 18080 for the Electron renderer
- allow same-origin loopback browsers while preserving sandbox token isolation


RussellLuo:
@jialudev Please help review this.

jialudev (approved):
LGTM

PR #424, opened 2026-07-30
feat(api): stream agent session responses

wanghaojie124:


RussellLuo:

internal/runtime/codex/event_sink.go
```diff
@@ -98,26 +101,48 @@ func (s *sessionSubscription) deliver(event SessionEvent) {
 }
 
 func (s *sessionSubscription) sendReliable(event SessionEvent) {
+	s.reliableMu.Lock()
+	s.reliableQueue = append(s.reliableQueue, event)
```
xxx7xxxx, line 118: This removes publisher backpressure by making every reliable subscriber queue unbounded, but the direct streaming path leaves the existing Codex bridge subscriber idle because its IM message is marked as a control record. Every direct turn is therefore copied into that idle channel and queue; I reproduced 192 retained stale events after the active stream had drained. Repeated streams can grow memory without bound, and the next ordinary turn in the same room can consume the old deltas and completion. Please avoid fan-out to the inactive bridge, use a session-scoped or single consumer, or otherwise give this queue a bounded lifecycle.

internal/runtime/codex/appserver_events.go
```diff
@@ -194,6 +206,40 @@ func (m *appServerManager) handleRawErrorNotification(runtimeID string, live *li
...
+			}
+			if phase == "" {
+				phase = "unknown"
+			}
+			payload["phase"] = phase
+			if phase == "final_answer" {
+				live.markStreamedAgentMessage(itemID)
+				live.markStreamedAgentThread(threadID)
+			}
+			m.publishAppServerEvent(SessionEvent{
```
xxx7xxxx: Anonymous session responses do not support interactive user-input requests, but this publishes final-answer deltas before `decodeAndPublishStructuredAssistantOutput` can remove private `:::csgclaw-output` records. I reproduced a `request_user_input` record verbatim inside `response.output_text.delta`. Because the session SSE path ignores structured events, callers receive the unsupported control record as answer text. Please filter or buffer the structured suffix incrementally, or stream from sanitized renderer output, so these records never enter the public response text.

internal/api/agent_sessions.go
```diff
@@ -216,50 +314,254 @@ func (h *Handler) createAgentSessionResponse(w http.ResponseWriter, r *http.Requ
...
+	runtimeSessionID string,
+) (string, error) {
+	if h.sessionEventSource == nil || runtimeEvents == nil {
+		return "", errSessionRuntimeEnded
+	}
+	ctx, cancel := context.WithTimeout(ctx, agentSessionResponseTimeout)
+	defer cancel()
+	promptDone := make(chan error, 1)
+	go func() {
+		promptDone <- h.sessionEventSource.Prompt(ctx, runtimeID, runtimeSessionID, prompt)
```
RussellLuo, line 365: Non-blocking follow-up for the initial text-streaming MVP: this direct `Prompt` path bypasses the codexbridge worker, while the loop below only handles text deltas and terminal events. `ActionRequest`, `UserInputRequest`, and structured-output events therefore never get bound or delivered to this room: approval/native-input turns can wait until the five-minute timeout, while structured `request_user_input` can finish without activating the question. Please track routing these events through the bridge's activity/user-input handling; a smaller fail-fast guard for unsupported interactive turns would also avoid the timeout.

internal/runtime/codex/appserver_events.go
```diff
@@ -380,11 +556,16 @@ func (m *appServerManager) handleLegacyAppServerEvent(runtimeID string, live *li
 	msgType := appServerString(params, "type")
 	switch msgType {
 	case "task_started":
+		live.clearStreamedAgentThread(threadID)
 		live.notifyAppServerTurn(threadID, appServerTurnResult{activity: "status:running", started: true})
 	case "agent_message":
 		text := appServerString(params, "message")
 		text = m.decodeAndPublishStructuredAssistantOutput(runtimeID, threadID, appServerString(params, "id"), text, live)
 		if text != "" {
+			if live.hasStreamedAgentThread(threadID) {
```
RussellLuo, line 610: This should be fixed before merge: the new guard suppresses the legacy `codex/event` final after typed final-answer deltas, but the other supported legacy final route, `codex/response_item`, does not check `hasStreamedAgentThread`. On a mixed legacy connection, the deltas are followed by the full response item and streaming clients receive the answer twice (for example, `hello worldhello world`). Apply the same streamed-thread guard in `handleLegacyResponseItemEvent` and add a mixed-protocol regression test.

internal/api/agent_sessions.go
```diff
@@ -216,50 +314,254 @@ func (h *Handler) createAgentSessionResponse(w http.ResponseWriter, r *http.Requ
...
+	}()
+	promptReturned := false
+	runtimeDone := false
+	finalize := func() (string, error) {
+		finalText := eventStream.text.String()
+		if strings.TrimSpace(finalText) != "" {
+			_, _ = h.im.DeliverMessage(im.DeliverMessageRequest{
```
RussellLuo: Non-blocking follow-up: the direct streaming path discards the `DeliverMessage` error and still emits `response.completed`. If IM persistence fails, the caller sees success while the room is missing the final assistant audit message. Please propagate this error to `response.failed` (or otherwise surface it) instead of silently reporting completion.

internal/runtime/codex/event_sink.go
wanghaojie124: Fixed in `d2e56b0a`. Event subscriptions are now scoped by runtime and session, and the Codex bridge subscribes per turn and cancels the subscription when the turn ends. This prevents direct streams from fanning reliable events into an idle long-lived bridge queue. Added session-isolation and lifecycle regression coverage.

internal/runtime/codex/appserver_events.go
wanghaojie124: Fixed in `d2e56b0a`. Final-answer deltas now pass through an incremental structured-output sanitizer before publication, so private `:::csgclaw-output` records never enter public SSE text. The sanitizer preserves ordinary streaming (including UTF-8 text) and buffers only a possible control-record boundary. Added chunked-record and Unicode regression tests.

internal/api/agent_sessions.go
wanghaojie124: Addressed in `d2e56b0a` with the fail-fast option for the initial text-streaming MVP. Anonymous direct streams now fail immediately with `response.failed` when they receive an approval request, native user-input request, or structured `request_user_input`, and cancel the prompt instead of waiting for the five-minute timeout. Added coverage for all three event paths.

internal/runtime/codex/appserver_events.go
wanghaojie124: Fixed in `d2e56b0a`. `handleLegacyResponseItemEvent` now applies the same `hasStreamedAgentThread` guard before publishing an assistant message, while still notifying turn progress. Added a mixed typed-delta/`codex/response_item` regression test that verifies the final answer is emitted only once.

internal/api/agent_sessions.go
wanghaojie124: Fixed in `d2e56b0a`. The direct streaming path now propagates `DeliverMessage` persistence failures, causing the SSE response to end with `response.failed` instead of `response.completed`. Added a regression test that removes the audit room before finalization and verifies failure is surfaced.

RussellLuo:
LGTM

RussellLuo:
If this is going to be merged, could you add a brief description to the commit message?

PR #426, opened 2026-07-31
feat(agent): separate runtime readiness from lifecycle

GatewayJ:
## Summary
- separate durable runtime lifecycle from transient gateway readiness
- refresh expired degraded readiness observations without probing every roster entry
- surface readiness and configured-startup state in the workspace
- https://git-devops.opencsg.com/product/agentichub/requirements/-/issues/3082
## Validation
- `go test -race ./internal/agent`
- targeted runtime, sandbox, API, and frontend tests
- frontend typecheck, lint, and format check

## Impact
- adds `runtime.availability` and `runtime.startup_pending` response fields; callers should use availability for gateway health


internal/agent/service.go
```diff
@@ -2006,16 +2016,71 @@ func (s *Service) ListContext(ctx context.Context) []Agent {
...
+		go func() {
+			defer group.Done()
+			for idx := range indices {
+				current := s.hydrateAgentStatus(ctx, agents[idx])
+				// A readiness probe is intentionally not part of the normal
+				// roster path. The only exception is an observation that was
+				// already degraded and has reached its TTL: recheck it here so
+				// an ongoing gateway failure remains visible without probing
+				// every running agent on every list request.
+				agents[idx] = s.refreshExpiredDegradedRuntimeAvailability(ctx, current)
```
wanghaojie124: The availability cache is ephemeral, so after every server restart all running gateways begin as `unknown`. This path only probes entries that were already recorded as `degraded`, which means a running container whose gateway is dead will remain `unknown` indefinitely and the workspace renders it as online until someone opens that agent's detail endpoint. Please seed readiness for unknown running gateways during startup (or add another bounded/background refresh path) so a restart does not erase gateway failures from the roster.

internal/agent/service.go
```diff
@@ -2006,42 +2016,105 @@ func (s *Service) ListContext(ctx context.Context) []Agent {
...
 		}
 		if _, err := s.Start(ctx, live.ID); err != nil {
 			startErr = errors.Join(startErr, fmt.Errorf("%s: %w", live.Name, err))
 		}
 	}
+	// Readiness observations are intentionally not persisted. Prime the running
+	// gateway workers after a process restart so an already-dead gateway does
+	// not look healthy until a user opens its detail endpoint. This pass runs
+	// with bounded concurrency while the startup marker keeps the roster fresh.
+	s.primeUnknownGatewayRuntimeAvailability(ctx, runningGateways)
```
wanghaojie124, line 2117: This one-shot startup seed fixes the empty-cache case, but it still cannot discover failures that begin after startup. A `ready` observation expires to `unknown` after 30 seconds, and the roster only re-probes expired `degraded` entries; this function is never called again. Consequently a gateway that dies after being seeded healthy remains lifecycle-`running` and is rendered online indefinitely unless its detail endpoint is opened. Please add a bounded ongoing refresh path for expired ready/unknown gateways (for example a rate-limited background sampler) so new failures are observable.

internal/agent/runtime_state.go
```diff
@@ -308,6 +363,219 @@ func (s *Service) RuntimeView(ctx context.Context, id string) (RuntimeView, erro
...
+
+	workers := min(agentListRuntimeProbeConcurrency, len(candidates))
+	indices := make(chan int)
+	var group sync.WaitGroup
+	group.Add(workers)
+	for range workers {
+		go func() {
+			defer group.Done()
+			for idx := range indices {
+				_ = s.refreshRuntimeAvailability(ctx, candidates[idx])
```
wanghaojie124: These startup probes use the long-lived server context, unlike the detail/list handlers' 2-second timeout. `CheckReadiness` ultimately runs Docker operations with this context and adds no deadline of its own, so an unresponsive Docker control plane can block a worker forever, prevent `group.Wait()` from returning, and leave `startup_pending` set indefinitely. Please wrap each probe (or the whole priming pass) in a bounded timeout and add a regression test with a checker that waits for context cancellation.

wanghaojie124:
LGTM

PR #430, opened 2026-08-03
fix(agent): restart codex agents after config changes

wanghaojie124:
- Conversation mappings are restored only through hydratePersistedSession. The profile-change restart follows Runtime.Stop → Runtime.Start, where Start creates an empty ConversationSessions map and overwrites session.json. Therefore, the automatic restart can still lose all room-to-thread mappings.
- Conversation mapping snapshots are persisted after releasing live.mu. Concurrent updates may write snapshots out of order, allowing an older snapshot to overwrite newer state.
- The race-enabled tests pass, but there is no integration test covering conversation preservation across a complete Runtime.Stop → Runtime.Start cycle.

### DIFF
--- a/internal/agent/lifecycle.go
+++ b/internal/agent/lifecycle.go
@@ -18,7 +18,10 @@ func (s *Service) restartCodexRuntime(ctx context.Context, id string) (Agent, er
 		return Agent{}, err
 	}
 	defer release()
+	return s.restartCodexRuntimeLocked(ctx, id)
+}
 
+func (s *Service) restartCodexRuntimeLocked(ctx context.Context, id string) (Agent, error) {
 	got, ok := s.Agent(id)
 	if !ok {
 		return Agent{}, fmt.Errorf("agent %q not found", strings.TrimSpace(id))
@@ -35,6 +38,12 @@ func (s *Service) restartCodexRuntime(ctx context.Context, id string) (Agent, er
 	s.stopLifecycleAgent(got.ID)
 	if _, err := runtimeImpl.Stop(ctx, handle); err != nil {
 		return Agent{}, fmt.Errorf("stop codex agent for restart: %w", err)
+	}
+	if _, err := s.updateRuntimeState(got.ID, agentruntime.Info{
+		HandleID: strings.TrimSpace(got.BoxID),
+		State:    agentruntime.StateStopped,
+	}); err != nil {
+		return Agent{}, fmt.Errorf("save stopped codex agent state: %w", err)
 	}
 	if err := s.provisionRuntimeForAgent(ctx, runtimeImpl, got, ""); err != nil {
 		return Agent{}, fmt.Errorf("provision codex agent for restart: %w", err)
--- a/internal/agent/lifecycle.go
+++ b/internal/agent/lifecycle.go
@@ -4,9 +4,82 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	agentruntime "csgclaw/internal/runtime"
 )
+
+// restartCodexRuntime restarts the in-process Codex app-server without deleting
+// its runtime directory. In particular, persisted Codex threads and conversation
+// mappings survive profile changes such as reasoning effort updates.
+func (s *Service) restartCodexRuntime(ctx context.Context, id string) (Agent, error) {
+	ctx, release, err := s.acquireAgentLifecycle(ctx, id)
+	if err != nil {
+		return Agent{}, err
+	}
+	defer release()
+
+	got, ok := s.Agent(id)
+	if !ok {
+		return Agent{}, fmt.Errorf("agent %q not found", strings.TrimSpace(id))
+	}
+	if !strings.EqualFold(strings.TrimSpace(got.RuntimeKind), RuntimeKindCodex) {
+		return Agent{}, fmt.Errorf("agent %q is not a codex agent", got.ID)
+	}
+
+	runtimeImpl, err := s.runtimeForKind(RuntimeKindCodex)
+	if err != nil {
+		return Agent{}, err
+	}
+	handle := runtimeHandleForAgent(got)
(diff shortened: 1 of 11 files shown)


RussellLuo:

xxx7xxxx:
@wanghaojie124 Better to have bullet points for the summary

internal/agent/lifecycle.go
```diff
@@ -4,10 +4,92 @@ import (
...
+	if _, err := s.updateRuntimeState(got.ID, agentruntime.Info{
+		HandleID: strings.TrimSpace(got.BoxID),
+		State:    agentruntime.StateStopped,
+	}); err != nil {
+		return Agent{}, fmt.Errorf("save stopped codex agent state: %w", err)
+	}
+	if err := s.provisionRuntimeForAgent(ctx, runtimeImpl, got, ""); err != nil {
+		return Agent{}, fmt.Errorf("provision codex agent for restart: %w", err)
+	}
+	state, err := runtimeImpl.Start(ctx, handle)
```
xxx7xxxx, line 51: This restart enters `Runtime.Start`, but that path still creates `SessionSpec` without reading the new `conversation_sessions` field. `ensureSession` then persists the restarted session with an empty map, so every automatic profile/MCP restart erases all room-to-thread mappings. I reproduced this with `Stop -> Start`: the saved `room-1` mapping was absent from the restart spec. Load the mapping in the normal `Start` path and add a stop/start regression test.

internal/runtime/codex/appserver_manager.go
```diff
@@ -126,14 +126,18 @@ func (m *appServerManager) Start(ctx context.Context, spec SessionSpec) (*Sessio
...
+	}
 	live := &liveSession{
 		cmd:                   cmd,
 		stdin:                 stdin,
 		stderr:                stderrFile,
 		done:                  make(chan struct{}),
 		spec:                  spec,
 		appClient:             appClient,
-		conversationSessions:  make(map[string]string),
+		conversationSessions:  conversationSessions,
```
xxx7xxxx, line 140: These persisted IDs are copied into memory but never resumed into the fresh app-server process. `EnsureSession` returns a cached ID immediately, then `Prompt` sends `turn/start` without a preceding `thread/resume`; a protocol-faithful reproduction fails with `thread not loaded` (`-32000`). Lazily resume restored room threads before returning them, handle resume fallback by updating persistence, and test through `Prompt` rather than only checking the cached ID.

internal/runtime/codex/appserver_manager.go
```diff
@@ -377,11 +382,21 @@ func (m *appServerManager) EnsureSession(ctx context.Context, handle SessionHand
 	}
 
 	live.mu.Lock()
-	defer live.mu.Unlock()
 	if existing := strings.TrimSpace(live.conversationSessions[conversationKey]); existing != "" {
+		live.mu.Unlock()
 		return existing, nil
 	}
 	live.conversationSessions[conversationKey] = threadID
+	conversations := cloneConversationSessions(live.conversationSessions)
```
xxx7xxxx, line 428: The snapshot is persisted after releasing `live.mu`, so overlapping create/reset operations can complete writes out of order. I reproduced a newer `{room-1, room-2}` snapshot being written first and then overwritten by a delayed `{room-1}` snapshot, which loses `room-2` after restart. Serialize persistence per live session or add versioning, and cover concurrent updates.

internal/agent/service_profiles.go
```diff
@@ -503,15 +507,19 @@ func (s *Service) update(ctx context.Context, id string, req UpdateRequest) (Age
 			}
 		}
 	}
-	if restartRequired && runtimeRunning && !isGatewayRuntimeKind(runtimeKind) {
+	if restartRequired && runtimeRunning && strings.EqualFold(runtimeKind, RuntimeKindCodex) {
+		if _, err := s.restartCodexRuntime(ctx, id); err != nil {
```
xxx7xxxx, line 511: This new restart path makes the existing Codex MCP API subtest terminate its own test process: `apiFakeCodexManager` stores `os.Getpid()` and its `Stop` is a no-op, so `Runtime.Stop` sends SIGINT to the package. `go test ./internal/api -run '^TestHandleAgentsMCPServersClosedLoopForSupportedRuntimes$/^codex$'` fails deterministically with `signal: interrupt`. Update the fixture to model a safely stoppable process and assert the restart behavior.

internal/agent/lifecycle.go
wanghaojie124: Fixed in 52def89d. Runtime.Start now reads the persisted session metadata and passes ConversationSessions into the new SessionSpec. I also added a Stop -> Start regression test that verifies the room-to-thread mapping reaches the restarted session.

internal/runtime/codex/appserver_manager.go
wanghaojie124: Fixed in 52def89d. Restored room threads are now lazily resumed before EnsureSession returns them. If resume fails, a new thread is created and the replacement mapping is persisted. The regression test now sends a Prompt through the restored thread to verify it is loaded in the fresh app-server process.

internal/runtime/codex/appserver_manager.go
wanghaojie124: Fixed in 52def89d. Conversation mapping mutations and persistence are serialized per live session, preventing older snapshots from being written after newer ones. Added a concurrent persistence regression test and verified the Codex runtime tests with the race detector.

internal/agent/service_profiles.go
wanghaojie124: Fixed in 52def89d. The API fixture now launches a dedicated test child process and tracks/stops it through the fake manager instead of using the package test PID. The exact Codex MCP subtest now passes, including with the race detector.

xxx7xxxx (approved):

PR #435, opened 2026-08-04
feat(codex): bundle Codex CLI with CSGClaw

GatewayJ:
## Summary

- Bundle the matching Codex CLI into local builds and official artifacts through one platform mapping.
- Resolve Codex only from the CSGClaw bundle and reject invalid prepared upgrade bundles.
- Allow marker-bearing legacy bundles that lack Codex to auto-upgrade to the new layout.
- Harden PowerShell Codex downloads: suppress slow progress rendering, retry transient failures, and promote only a completed temporary download.

## Validation

- `go test ./...`
- `git diff --check`
- `bash -n scripts/fetch-codex-cli.sh`

## Impact

- Codex app-server continues to run from the CSGClaw bundle; sandbox runtime assets remain limited to `csgclaw-cli`.
- Existing official installs without a bundled Codex CLI can use in-app upgrade to migrate.
- Native Linux/macOS packaging continues to use the existing Bash `curl` downloader; only the PowerShell downloader changed.


internal/upgrade/download.go
```diff
@@ -295,6 +296,9 @@ func inspectBundleDir(bundleDir string) (BundleLayout, error) {
 	if err != nil {
 		return BundleLayout{}, err
 	}
+	if _, err := requiredBundleExecutable(bundleDir, "codex"); err != nil {
```
jialudev: Please keep pre-bundled official installations upgradeable. `officialInstallRoot()` reuses `validateBundleDir()`, so this new required `codex` check rejects every currently released marker-bearing bundle that has only `bin/csgclaw`; `isLegacyOfficialInstallRoot()` also rejects it because the marker exists. I reproduced `AutoUpgradeSupport("v0.4.5")` returning `reason="not_official_bundle"`. Please separate validation for the incoming bundle from compatibility validation of the current install, or explicitly accept the previous marked layout for one-way migration.

internal/codexcli/bundled.go
```diff
@@ -0,0 +1,117 @@
...
+	Stat           func(string) (os.FileInfo, error)
+}
+
+func (l BundleLocator) Locate() (string, error) {
+	executable, err := l.executablePath()
+	if err != nil {
+		return "", fmt.Errorf("resolve CSGClaw executable for bundled Codex CLI: %w", err)
+	}
+
+	candidate := filepath.Join(filepath.Dir(l.executablePathCandidate(executable)), l.binaryName())
```
jialudev, line 45: Please cover the documented source-build path before requiring this sibling binary. `make build-server-bin` and `make run` only produce `bin/csgclaw` and `bin/csgclaw-cli`, so after a clean build there is no `bin/codex`; the fixed Codex manager then fails to start. Please either stage Codex in the local build targets or keep an explicit development-only resolution path while retaining strict bundled resolution for official releases.

scripts/fetch-codex-cli.sh
```diff
@@ -0,0 +1,75 @@
...
+
+if [ "$#" -ne 3 ]; then
+  echo "usage: $0 <goos> <goarch> <output-dir>" >&2
+  exit 1
+fi
+
+GOOS_TARGET="$1"
+GOARCH_TARGET="$2"
+OUTPUT_DIR="$3"
+CODEX_CLI_DOWNLOAD_BASE_URL="${CODEX_CLI_DOWNLOAD_BASE_URL:-https://csgclaw.opencsg.com/codex-cli/latest}"
```
jialudev, line 12: Please pin the bundled Codex version and checksum instead of building releases from `/latest`. Release and Desktop matrix jobs download independently, so the same CSGClaw release can contain different Codex versions if `latest` changes, and rebuilding the same tag is not reproducible. An explicit `CODEX_CLI_VERSION` plus per-platform checksums, propagated through the shell, PowerShell, and Docker paths, would preserve the tested matching-version contract.

internal/codexcli/bundled_test.go
```diff
@@ -0,0 +1,101 @@
...
+func TestBundleLocatorLocateUsesCodexBesideExecutable(t *testing.T) {
+	root := t.TempDir()
+	executable := writeBundledExecutable(t, filepath.Join(root, "bin", "csgclaw"))
+	codex := writeBundledExecutable(t, filepath.Join(root, "bin", "codex"))
+
+	got, err := (BundleLocator{ExecutablePath: func() (string, error) { return executable, nil }}).Locate()
+	if err != nil {
+		t.Fatalf("Locate() error = %v", err)
+	}
+	if got != codex {
```
jialudev: These tests fail on macOS because `t.TempDir()` returns `/var/folders/...` while `EvalSymlinks` canonicalizes the executable to `/private/var/folders/...`; `Locate()` returns the latter and the direct string comparison fails. Four new tests fail this way. Please canonicalize the expected path or compare file identity with `os.SameFile`.

jialudev (dismissed):
Requesting changes for four release and development-path regressions: preserve upgrades from existing official bundles, keep the documented local build/run path functional, pin the bundled Codex artifact, and make the new locator tests pass on macOS.

jialudev:
LGTM

jialudev (approved):

PR #440, opened 2026-08-05
ci(desktop): publish release installers to OSS

jialudev:
- Publish signed macOS and Windows desktop installers to immutable OSS version paths after the GitHub Release succeeds.
- Generate beta and release channel manifests from SemVer tags, and mark prerelease tags as GitHub Pre-releases.
- Reuse the desktop artifact contract across local Make targets and CI while loading OSS credentials from the shared `oss-publish` environment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,6 @@ jobs:
       desktop_matrix: ${{ steps.set-matrix.outputs.desktop_matrix }}
       version: ${{ steps.release_meta.outputs.version }}
       release_channel: ${{ steps.release_meta.outputs.release_channel }}
-      oss_environment: ${{ steps.release_meta.outputs.oss_environment }}
       oss_manifest_url: ${{ steps.release_meta.outputs.oss_manifest_url }}
       commit: ${{ steps.release_meta.outputs.commit }}
       build_time: ${{ steps.release_meta.outputs.build_time }}
@@ -89,14 +88,8 @@ jobs:
             version="${GITHUB_REF_NAME}"
           fi
           release_channel="$(node scripts/desktop-release-artifacts.mjs channel "${version}")"
-          case "${release_channel}" in
-            beta) oss_environment="stg" ;;
-            release) oss_environment="prd" ;;
-            *) echo "unexpected release channel: ${release_channel}" >&2; exit 1 ;;
-          esac
           echo "version=${version}" >> "$GITHUB_OUTPUT"
           echo "release_channel=${release_channel}" >> "$GITHUB_OUTPUT"
-          echo "oss_environment=${oss_environment}" >> "$GITHUB_OUTPUT"
           echo "oss_manifest_url=https://opencsg-public-resource.oss-cn-beijing.aliyuncs.com/csgclaw-desktop/channels/${release_channel}/downloads.json" >> "$GITHUB_OUTPUT"
           echo "commit=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
           echo "build_time=$(date -u +%Y-%m-%dT%H:%M:%SZ)" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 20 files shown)


scripts/desktop-oss-release.mjs
```diff
@@ -0,0 +1,449 @@
...
+  for (const [existingVersion, value] of Object.entries(previous?.versions || {})) {
+    if (existingVersion !== version) {
+      versions[existingVersion] = value;
+    }
+  }
+
+  const manifest = {
+    schema_version: 1,
+    channel,
+    latest: version,
```
wanghaojie124, line 111: This unconditionally makes the version being published `latest`, even when the remote manifest already points to a newer SemVer. Re-running an older tag (or manually dispatching one) would therefore roll the public channel back to that older installer. Please compare against the existing `latest` and reject older versions (with an explicit override if rollback is intentional), and add a regression test for publishing an older version into a newer manifest.

scripts/desktop-oss-release.mjs
```diff
@@ -0,0 +1,449 @@
...
+    ...process.env,
+    OSS_ACCESS_KEY_ID: accessKeyID,
+    OSS_ACCESS_KEY_SECRET: accessKeySecret,
+    OSS_REGION: region,
+    OSS_ENDPOINT: endpoint,
+  };
+
+  runCommand("ossutil", ["version"], ossEnvironment);
+  const manifestURL = `${publicBaseURL}/channels/${context.channel}/downloads.json`;
+  const currentResponse = await fetch(`${manifestURL}?current=${Date.now()}`, { cache: "no-store" });
```
wanghaojie124, line 346: This is an unprotected read-modify-write of the shared channel manifest. If two releases for the same channel overlap, both can read the same old manifest and the last upload drops the other release from `versions` (and may also leave the older run as `latest`). Please serialize publishing per channel in the workflow and/or use an ETag/conditional write with a retrying merge.

jialudev: Fixed in fec974e3: serialize queued OSS publish jobs per channel without canceling them; platform builds and GitHub Releases remain parallel.

scripts/desktop-oss-release.mjs
jialudev: Fixed in b8e18c27: reject publishing a version older than the current latest and cover the rollback case with a regression test.

.github/workflows/release.yml
```diff
@@ -376,4 +388,72 @@ jobs:
...
           tag_name: ${{ needs.prepare.outputs.version }}
+          prerelease: ${{ needs.prepare.outputs.release_channel == 'beta' }}
           files: dist/*
+
+  publish-desktop-oss:
+    needs: [prepare, merge-and-release]
+    concurrency:
+      group: desktop-oss-${{ needs.prepare.outputs.release_channel }}
+      cancel-in-progress: false
+      queue: max
```
wanghaojie124, line 399: `queue` is not a supported key in a GitHub Actions `concurrency` block; only `group` and `cancel-in-progress` are accepted. This makes the release workflow invalid, so tag-triggered releases will not start. Please remove `queue: max`; `cancel-in-progress: false` already causes pending runs in this concurrency group to wait (noting that GitHub does not guarantee queue ordering).

jialudev: Thanks for flagging this. `queue: max` is supported by GitHub Actions as of May 7, 2026. It can be used when `cancel-in-progress` is `false` or omitted, and is required here so multiple channel releases can wait without replacing an existing pending run. See the [GitHub Changelog announcement](https://github.blog/changelog/2026-05-07-github-actions-concurrency-groups-now-allow-larger-queues/) and the [official concurrency documentation](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency).

wanghaojie124:
LGTM

PR #443, opened 2026-08-05
feat(hub): add community template publishing and deployment

wanghaojie124:
Summary:
- add template-only and publish-and-deploy community workflows
- aggregate Agentic and CSGClaw official templates with namespace/name IDs
- create private CSGClaw instances using published repository paths
- resolve unauthenticated template access from CSGHUB_BASE_URL and CSGHUB_USER_TOKEN
- update CLI, Web UI, manager skill, documentation, and test coverage


internal/template/remote_store.go
```diff
@@ -456,8 +554,60 @@ func (s *RemoteStore) fetchBlob(ctx context.Context, id, filePath, branch string
...
+		Name:          normalized.Name,
+		Nickname:      normalized.Name,
+		Description:   normalized.Description,
+		Private:       false,
+		DefaultBranch: "main",
+		Type:          "template",
+		CodeFile:      strings.TrimSpace(upload.Data.UUID),
+	}
+	var created remoteCodeResponse
+	if err := s.postJSON(ctx, s.codesURL(), request, &created); err != nil {
```
xxx7xxxx, line 607: This call only creates a Code repository. The current CSGHub Code handler forces the repository type to `CodeRepo`, so the submitted `type: "template"` value is ignored, and it does not create an `/agent/templates` record. Because `List` discovers personal community templates only through `/agent/templates` (plus the separate `Agentic` organization), a template-only publish under `<user>/<name>` succeeds but disappears after refresh and is not discoverable. Please register the created repository through the agent-template API using `metadata.repo_path`, or list published repositories from a source that actually includes them, and cover publish-then-list end to end.

internal/api/handler.go
```diff
@@ -1567,27 +1568,78 @@ func (h *Handler) handleHubTemplates(w http.ResponseWriter, r *http.Request) {
...
 			}
 			http.Error(w, err.Error(), status)
 			return
 		}
-		spec.Registry = req.Registry
-		item, err := hubSvc.Publish(r.Context(), spec)
-		if err != nil {
-			http.Error(w, err.Error(), http.StatusBadGateway)
-			return
+		if req.Deploy {
```
xxx7xxxx, line 1623: `publish and deploy` is not retry-safe here. The repository has already been created before the instance request runs; if instance creation times out or returns an error, this returns 502 while leaving the public template behind. Retrying then fails on the duplicate repository before deployment is attempted, so a transient second-step failure leaves the workflow stuck. Make this operation resumable/idempotent, for example by reusing the existing published repository on retry, or add a compensating cleanup before reporting failure, with a regression test for a failed instance request followed by retry.

internal/template/service.go
```diff
@@ -226,7 +229,40 @@ func (s *Service) Publish(ctx context.Context, spec PublishSpec) (Template, erro
...
+	if workspace.Temporary && strings.TrimSpace(workspace.Path) != "" {
+		defer func() { _ = os.RemoveAll(workspace.Path) }()
+	}
+	var mcpServers map[string]any
+	if raw := strings.TrimSpace(workspace.MCPServersJSON); raw != "" {
+		if err := json.Unmarshal([]byte(raw), &mcpServers); err != nil {
+			return Template{}, fmt.Errorf("decode hub template MCP servers: %w", err)
+		}
+	}
+	return s.Publish(ctx, PublishSpec{
```
xxx7xxxx, line 253: This template-copy path drops `item.ImageEnv`: `PublishSpec` has no corresponding field and `writeManifest` therefore emits no `[[image.env]]` entries. A local template that requires `GITLAB_TOKEN`, for example, is published without its required secret/default contract, so consumers are no longer prompted for or given the values needed to run it. Carry the image environment contracts through `PublishSpec` and manifest serialization, and add an archive round-trip test.

web/app/src/pages/AgentPage/components/AgentDetailPane/AgentDetailPane.tsx
```diff
@@ -276,14 +285,39 @@ export const AgentDetailPane = forwardRef<AgentDetailPaneHandle, AgentDetailPane
   const profile = agentProfileConfig(item);
   const provider = item.provider || profile?.provider || providerNameForProviderID(profile?.model_provider_id || "");
   const runtimeKind = agentRuntimeKind(item);
-  const canPublish = runtimeKind === "picoclaw_sandbox" || runtimeKind === "openclaw_sandbox";
+  const canPublishLocal = runtimeKind === "codex" || runtimeKind === "openclaw_sandbox";
```
xxx7xxxx, line 288: This removes the existing local-template action for `picoclaw_sandbox` agents. The backend still supports PicoClaw snapshots (`HubPublishSpec` and the local store tests cover them), and before this PR the UI explicitly allowed PicoClaw publishing. As written, existing PicoClaw users lose the only Web UI path to save a local template. Keep `picoclaw_sandbox` in `canPublishLocal` and add a visibility regression test; only the community actions need to remain Codex-only.

internal/template/remote_store.go
```diff
@@ -96,14 +140,46 @@ func NewRemoteStore(baseURL, token string) *RemoteStore {
...
+	var listErrs []error
+	var organizationTemplates remoteCodeListResponse
+	if err := s.getJSON(ctx, s.templatesURL(), &organizationTemplates); err != nil {
+		listErrs = append(listErrs, err)
+	} else {
+		repositories = appendRemoteTemplateRepositories(repositories, seen, organizationTemplates.Data)
+	}
+
+	var agentTemplates remoteAgentTemplateListResponse
+	if err := s.getJSON(ctx, s.agentTemplatesURL(), &agentTemplates); err != nil {
```
xxx7xxxx: (nit): This reads only one `/agent/templates` response and ignores `Total`. CSGHub defaults this endpoint to `per=20&page=1`, so after the catalog exceeds 20 CSGClaw templates, every later page silently disappears from the merged Hub list. Iterate pages until all `Total` entries are collected (or use and validate an explicit supported page size), with a test containing more than one page.

internal/template/embed/manager/codex/skills/agent-creator/SKILL.md
```diff
@@ -15,8 +15,8 @@ Before running **any** `csgclaw-cli participant create --type agent --bind creat
 
 1. Read this skill first.
 2. Run `csgclaw-cli --output json template list` and pick a template (do not skip even if the user named a capability like GitLab).
-3. Run `csgclaw-cli --output json template get <template-id>`.
-4. Create with `--from-template` and required `--env` values.
+3. Run `csgclaw-cli --output json template get <template-id>` using the exact ID returned by `template list`.
+4. Create with that exact ID in `--from-template` and add required `--env` values. Remote template IDs are absolute URLs such as `https://hub.opencsg.com/namespace/name`; do not rewrite them as `official.namespace/name`.
```
xxx7xxxx: (nit): This contradicts the new API/CLI contract and the operating rule below: remote IDs returned by `template list` are `<namespace>/<name>`, not absolute URLs. The URL example is rejected by `normalizeRemoteTemplateID`, so a manager following this mandatory workflow can fail `template get` and worker creation. Use the exact `<namespace>/<name>` form consistently here, in the workflow text, and in the command example.

internal/template/remote_store.go
wanghaojie124: Thanks for raising this. Template registration is handled on the CSGHub side: after the Code repository is pushed, a Git hook automatically registers it as an agent template. CSGClaw then reads the complete catalog from the CSGHub `/agent/templates` API. Because registration is owned by that CSGHub workflow, CSGClaw should not create a second template record here, so we are keeping this publish path unchanged.

internal/api/handler.go
wanghaojie124: Thanks. The current product behavior is intentionally non-transactional: publishing and deployment are separate steps, and if deployment fails we return the concrete deployment error to the user while leaving the successfully published template available. Deployment failures in this flow are generally not transient, so automatically retrying or deleting the published repository would hide the useful failure state and could discard a successful publish. We are keeping the direct error-return behavior in this version.

internal/template/service.go
wanghaojie124: Thanks for pointing this out. The template-creation/configuration entry point is not currently implemented in CSGClaw, and the initial publish-and-deploy flow does not expose environment-variable configuration. For this version we are intentionally limiting support to templates that do not require additional image environment variables. Carrying and configuring `image.env` will be handled together with the future template-creation/configuration flow, so we are leaving this out of the current scope.

web/app/src/pages/AgentPage/components/AgentDetailPane/AgentDetailPane.tsx
wanghaojie124: Thanks. PicoClaw is no longer maintained, and this community publishing/deployment work is not being developed for the PicoClaw runtime. The UI restriction is therefore intentional, and we are not restoring or extending the PicoClaw publishing path in this change.

internal/template/remote_store.go
wanghaojie124: Fixed in `bd9e1e3f`. The CSGHub agent-template request now sends explicit `page` and `per=20` parameters and continues fetching pages until all entries reported by `Total` have been collected. I also added a regression test with 21 templates that verifies both pages are requested and all templates are returned.

internal/template/embed/manager/codex/skills/agent-creator/SKILL.md
wanghaojie124: Fixed in `bd9e1e3f`. The routing gate, workflow guidance, and command example now consistently use the exact `<namespace>/<name>` form returned by `template list`, including `Agentic/gitlab-assistant`, with no absolute-URL example.

GatewayJ:
模板地址使用环境变量拼接 考虑使用CSGHUB_API_BASE_URL 而不是新增CSGHUB_BASE_URL?

web/app/src/pages/AgentPage/components/AgentDetailPane/AgentDetailPane.tsx
xxx7xxxx: Ok

wanghaojie124:
@GatewayJ 已采纳，提交 `8eb40046` 已移除新增的 `CSGHUB_BASE_URL`，模板 Hub 地址统一复用现有的 `CSGHUB_API_BASE_URL`。相关 API 测试以及中英文配置文档也已同步更新。

GatewayJ:
lgtm

PR #445, opened 2026-08-05
fix(web): align attachments and gate unsupported downloads

youngbeom-shin:
## Summary

- right-align attachments in messages sent by the current user
- truncate sent attachment names like composer cards while keeping file extensions visible
- open the attachment preview dialog for unsupported file types instead of downloading immediately
- let users explicitly choose whether to download after seeing the unavailable-preview state


youngbeom-shin:
before：
image

after：
截屏2026-08-05 16 57 00

youngbeom-shin:
@jialudev please help review this alignment bug fix.

jialudev (approved):
LGTM

PR #452, opened 2026-08-06
fix(codex): allow active turns to run past 30 minutes

xxx7xxxx:
## Summary

- Remove the fixed 30-minute duration cap from active Codex turns.
- Keep inactivity and cancellation safeguards with regression coverage.


RussellLuo:
LGTM

PR #458, opened 2026-08-07
fix(web): localize auth network errors

youngbeom-shin:
## Summary

- Map browser-level authentication network failures to the existing localized UI messages.
- Show the predefined login-status error when `/api/v1/auth/status` cannot be reached.
- Keep specific non-network authentication errors unchanged.


jialudev (approved):
LGTM

PR #459, opened 2026-08-07
fix(web): improve model loading error UI

youngbeom-shin:
## Summary

- show model-loading failures in a full-width alert instead of expanding one form field
- add a retry action and collapsible error details while retaining the saved model selection
- keep the four model controls full-width with responsive four-, two-, and one-column layouts
- add localized Chinese and English copy plus focused component coverage


youngbeom-shin:
https://github.com/user-attachments/assets/1869b499-bf5a-45e5-b74c-f4dec81b1558

jialudev (approved):
LGTM

PR #465, opened 2026-08-10
fix(web): hide unsupported directory picker action

wanghaojie124:


internal/api/directory_picker.go
```diff
@@ -4,13 +4,99 @@ import (
...
+	default:
+		return false
+	}
+}
+
+func linuxDisplayAvailable(getenv func(string) string, connect displayConnector) bool {
+	waylandDisplay := strings.TrimSpace(getenv("WAYLAND_DISPLAY"))
+	if waylandDisplay != "" {
+		runtimeDir := strings.TrimSpace(getenv("XDG_RUNTIME_DIR"))
+		if runtimeDir != "" && connect("unix", filepath.Join(runtimeDir, waylandDisplay)) {
```
xxx7xxxx: (nit): This treats every `WAYLAND_DISPLAY` value as relative to `XDG_RUNTIME_DIR`, although Wayland also permits an absolute socket path. The X11 parser below likewise treats valid protocol-qualified displays such as `unix/:0` or `tcp/host:0` as literal hosts. Those valid sessions are therefore reported as unavailable and the UI hides a picker that can run. Please preserve absolute Wayland paths and handle the optional X11 protocol prefix, with regression cases for both forms.

web/app/src/components/business/ProfileControls/RuntimeOptionsFields.tsx
```diff
@@ -65,23 +67,25 @@ export function RuntimeOptionsFields({
...
-                  }
-                  onDraftChange({
-                    ...draft,
-                    runtime_options: setRuntimeOptionValue(draft.runtime_options, path, pickedPath),
-                  });
-                }}
-              >
-                {directoryPickerLabel(locale)}
-              </Button>
+              {directoryPickerAvailable ? (
```
xxx7xxxx: (nit): This flag only reports availability of the server-side picker, but `pickLocalDirectoryPath()` explicitly falls back to the renderer picker when the host endpoint is unavailable, and the existing frontend test asserts that contract. Hiding the button here makes that fallback unreachable whenever the server reports `false`. Please either include renderer capability in the visibility decision, or remove the fallback as an intentional contract change and update its tests.

internal/api/directory_picker.go
wanghaojie124: Fixed in `37e3835e`. Absolute `WAYLAND_DISPLAY` socket paths are now preserved, while relative values are still resolved against `XDG_RUNTIME_DIR`. The X11 parser now handles the optional `unix/` and `tcp/` protocol prefixes, with regression coverage for absolute Wayland sockets, `unix/:0`, and `tcp/host:0` forms.

web/app/src/components/business/ProfileControls/RuntimeOptionsFields.tsx
wanghaojie124: Fixed in `37e3835e`. Button visibility now combines the server-side capability with renderer availability (`showDirectoryPicker` or the CSGClaw Desktop runtime), so the existing renderer fallback remains reachable when the host picker is unavailable. Added helper and component regression tests for this behavior.

xxx7xxxx (approved):

PR #469, opened 2026-08-12
fix(web): hide activity from agent profiles

GatewayJ:
## Summary

- remove the Activity tab from public agent profile pages
- stop passing conversation rooms into the agent profile pane
- fall back to Profile when local storage contains the removed `activity` tab value

## Why

Agent profiles are public, while the activity feed only reflects CSGClaw channel activity. Hiding it avoids exposing channel-specific activity on a public profile and prevents an incomplete activity view from being presented as comprehensive.

## Impact

Conversation activity panels, working indicators, activity cards, and backend activity APIs are unchanged.

## Validation

- `vitest run tests/components/AgentActions.test.tsx src/components/business/ConversationPane/ConversationComposer.test.tsx src/pages/ConversationPage/components/ConversationActivityPanel/conversationActivity.test.ts`
- `tsc --noEmit`
- ESLint on changed files
- Prettier check on changed files

### DIFF
--- a/web/app/src/hooks/workspace/useAgentController.ts
+++ b/web/app/src/hooks/workspace/useAgentController.ts
@@ -2450,7 +2450,6 @@ export function useAgentController({
       saveError: agentPageError,
       notice: selectedAgentPageNotice?.message || "",
       noticeTone: selectedAgentPageNotice?.tone || "warning",
-      rooms,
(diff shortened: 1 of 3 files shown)


jialudev (approved):
LGTM

PR #470, opened 2026-08-12
feat(desktop): add update channels and turn notifications

jialudev:
- Add release and beta update channels for Server and desktop distributions.
- Notify users when agent turns finish, with configurable focus behavior and system permission guidance.
- Harden desktop signing, notarization, and CI publishing for complete macOS and Windows update feeds.


wanghaojie124 (commented):
High — `.github/workflows/release.yml:480`

`concurrency` only accepts `group` and `cancel-in-progress`; `queue` is not a supported GitHub Actions key. This makes the release workflow invalid, so tag and manual releases will not start. Please remove `queue: max`; `cancel-in-progress: false` already keeps an in-progress run from being cancelled (while concurrency itself allows at most one pending run per group).

wanghaojie124:
LGTM

web/app/src/hooks/workspace/useAgentTurnNotifications.ts
```diff
@@ -0,0 +1,256 @@
...
+    try {
+      const notification = new window.Notification(current.t("turnNotificationTitle", { agent: agentName }), {
+        body,
+        icon: "favicon.ico",
+        tag: `csgclaw-${input.eventKey}`,
+      });
+      notifiedEventKeysRef.current.add(input.eventKey);
+      notification.onclick = () => {
+        notification.close();
+        window.focus();
```
xxx7xxxx: When the Electron window is closed, `WindowManager` handles that by calling `BrowserWindow.hide()`. This renderer-side `window.focus()` does not show the hidden native window, so clicking a completion notification closes it and changes the route while CSGClaw remains invisible. Please route notification activation through a desktop IPC that calls the existing `showAndFocus()` path before selecting the room, while retaining `window.focus()` for regular browsers.

jialudev: Thanks for catching this. Fixed in 4f5f2e79: desktop notification activation now goes through a trusted IPC handler wired to AppLifecycle.show(), which reuses WindowManager.showAndFocus(). The handler is awaited before selecting the room, while regular browsers still use window.focus(). I also added a regression test that verifies the desktop window is shown before room selection.

PR #472, opened 2026-08-13
fix(team): keep manager selected in member dialog

GatewayJ:
## Summary

- keep the team manager selected and non-removable in create and edit dialogs
- restore existing team members while normalizing legacy and current agent IDs
- keep the team lead out of `member_agent_ids` when saving member changes

## Root cause

The edit dialog initialized its selection from `member_agent_ids` only, while the manager is stored separately as `lead_agent_id`. This made the manager appear unselected and allowed the UI selection state to remove it even though the backend continued treating it as the team lead.

## Impact

Team creation and member management now consistently reflect the backend lead/member model without changing the API contract or stored team data.

## Validation

- `pnpm --dir web/app exec vitest run` — 118 test files, 846 tests passed
- ESLint and Prettier checks for the changed files

## Related issue

- https://git-devops.opencsg.com/product/agentichub/requirements/-/issues/3298


jialudev (approved):
LGTM

PR #473, opened 2026-08-13
fix(skill-hub): open remote skills on the correct site

GatewayJ:
## What changed

- Build remote Skill detail links from the OpenCSG public site instead of the Hub API endpoint.
- Keep the existing `remote_url` API field for client compatibility.
- Use `https://opencsg.com` while signed out and the active OpenCSG login site while signed in.
- Recover production, stage, and custom site URLs from legacy login records that do not contain `opencsg_base_url`.
- Document the browser URL behavior and add regression coverage for each environment.

## Why

Remote Skill links were generated from the Hub API base URL. In production this produced `https://hub.opencsg.com/skills/...`, which redirects users to the OpenCSG home page instead of the Skill detail page.

The environment resolver is now the single source of truth for both Hub API and public site addresses, while preserving the existing API response contract.

Related requirement: https://git-devops.opencsg.com/product/agentichub/requirements/-/issues/3360

## Validation

- `go test ./internal/api -run '^(TestHandleRemoteSkillsUsesEffectiveOfficialHub|TestCurrentOpenCSGEnvironment|TestOpenCSGEnvironmentFromStatus)' -count=1`
- `go test ./internal/skill/remote -count=1`
- `pnpm --dir web/app exec vitest run tests/api/skills.test.ts tests/components/WorkspaceTabPanels.test.tsx`
- ESLint and Prettier checks for changed frontend tests
- `git diff --check`


jialudev (approved):
LGTM

PR #476, opened 2026-08-13
fix(skill): prefer Hub archive downloads

GatewayJ:
## Summary

- download remote Skills through the Hub archive endpoint when available
- retain the tree/blob fallback for older Hub deployments
- validate downloaded ZIPs before installation and use an archive-specific timeout

## Why

Remote Skill installation currently fetches every repository file through separate tree/blob calls. Large Skills can require thousands of requests and take several minutes even when the Hub exposes a repository archive endpoint. A malformed successful response also reached the local installer and surfaced as a 500.

## Impact

Supported Hub deployments install Skills with one archive request. Older deployments continue to work through the existing fallback. Invalid upstream archives now return 502, and large archive transfers have a five-minute deadline instead of the JSON request timeout.

## Validation

- `go test ./internal/skill/remote ./internal/skill/local -count=1`
- `go test ./internal/api -run '^(TestHandleSkillInstallFromOfficialHub|TestHandleRemoteSkillsUsesEffectiveOfficialHub|TestHandleSkillUploadRejectsDuplicate|TestHandleSkillUploadRejectsInvalidArchive)$' -count=1`
- `go vet ./internal/skill/remote ./internal/skill/local`
- validated the staging Hub archive endpoint and local install/replace flow

### DIFF
--- a/docs/tech/api.md
+++ b/docs/tech/api.md
@@ -152,7 +152,7 @@ The server keeps the current Hub catalog ordering and returns normalized Skill s
 
 ### `POST /api/v1/skills:install`
 
-Installs one remote Skill from the same effective OpenCSG Hub. Set `replace` to overwrite an existing local Skill with the same name.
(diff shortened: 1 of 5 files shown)


GatewayJ:
同步小助手的优化方案

wanghaojie124:
LGTM

PR #479, opened 2026-08-14
fix: clean up notification deletion and stale routes

xxx7xxxx:
## Summary

- Remove the orphaned local IM user and direct room when a notification participant is deleted.
- Reload once when a deployed lazy route chunk becomes stale, preventing the raw application error after updates.


wanghaojie124:
LGTM

PR #481, opened 2026-08-14
fix(auth): refresh rotated OpenCSG API keys

xxx7xxxx:
## Summary

- Refresh the built-in OpenCSG AI Gateway API key after an upstream 401 and retry once.
- Persist the refreshed key, serialize concurrent refreshes, and cover the behavior with tests.


jialudev (approved):
LGTM

PR #483, opened 2026-08-14
fix(agent): improve model and runtime failure handling

wanghaojie124:
- unify safe model validation and streaming error mappings, and filter unavailable non-text models
- preserve OpenCSG credential refresh while sanitizing bounded Responses SSE failures
- detect Docker availability without blocking agent lists or reusing stale runtime observations
- synchronize online, offline, and checking states across agent UI surfaces
- add backend and frontend regression coverage


internal/modelprovider/openai.go
```diff
@@ -70,37 +88,88 @@ func ListOpenAIModelsWithClient(ctx context.Context, client *http.Client, baseUR
...
+	}
 	for _, item := range payload.Data {
 		id := strings.TrimSpace(item.ID)
 		if id == "" {
 			continue
 		}
+		if item.Availability != nil && item.Availability.IsAvailable != nil && !*item.Availability.IsAvailable {
+			continue
+		}
+		if hasTaskMetadata && !taskSupportsTextGeneration(item.Task) {
```
xxx7xxxx: This treats a task-less entry as non-text whenever any sibling includes task metadata. The current AIGateway contract still emits task: "" for external models whose optional metadata.tasks is unset, so a mixed catalog hides otherwise usable custom chat models from CSGClaw's provider/model selectors. Please keep unknown-task entries and exclude only entries with an explicit non-text task, then add a mixed annotated/task-less catalog regression test.

internal/agent/runtime_state.go
```diff
@@ -479,11 +479,45 @@ func (s *Service) refreshRuntimeAvailabilityWithTimeout(ctx context.Context, a A
...
+	}
+	runtimeImpl, err := s.runtimeForKind(strings.TrimSpace(a.RuntimeKind))
+	if err != nil {
+		return a
+	}
+	if _, ok := runtimeImpl.(agentruntime.ReadinessChecker); !ok {
 		return a
 	}
-	return s.refreshRuntimeAvailabilityWithTimeout(ctx, a)
+	s.scheduleRuntimeAvailabilityRefresh(a)
```
xxx7xxxx, line 494: This schedules the first readiness observation after returning the list. A newly created or replaced sandbox agent therefore comes back as unknown with no expires_at; once useWorkspaceAgentsQuery is past its two-minute startup window, that shape schedules no refetch, and this goroutine updates only the server cache without emitting an event. The UI can remain Checking indefinitely until an unrelated refresh. Please give pending unknown observations a near expiry, publish or invalidate on completion, or keep the first probe bounded in the request, and cover creation after the startup window.

internal/modelprovider/upstream_error.go
```diff
@@ -0,0 +1,165 @@
...
+	case "invalid_request_error", "context_length_exceeded", "unsupported_feature", "unsupported_model", "model_task_mismatch":
+		return http.StatusBadRequest
+	default:
+		return http.StatusServiceUnavailable
+	}
+}
+
+// FriendlyUpstreamErrorMessage maps upstream status and error codes to text
+// that is safe and actionable for end users.
+func FriendlyUpstreamErrorMessage(status int, code string) string {
```
xxx7xxxx, line 125: All branches here, and friendlySandboxRuntimeMessage, return Chinese text from server-side API and SSE contracts. writeAgentOperationError then discards the stable code and sends that text directly, so the English UI and API or CLI consumers receive Chinese errors even though the product supports English. Please preserve a structured error code for client localization or make server messages locale-neutral, and cover the English locale path.

internal/modelprovider/openai.go
wanghaojie124: Fixed in 7bcac1ee. Task-less and empty-task entries are now preserved; only models with an explicit non-text task are filtered out. The mixed annotated/task-less catalog test now covers both empty and missing task metadata.

internal/agent/runtime_state.go
wanghaojie124: Fixed in 7bcac1ee. The workspace query now keeps polling at a short interval when a runtime availability observation is unknown and has no expires_at, including after the startup polling window. A regression test covers this pending-unknown shape. I also prevented an in-flight probe from restoring availability after its agent has been deleted.

internal/modelprovider/upstream_error.go
wanghaojie124: Fixed in 7bcac1ee. Agent operation errors now return a structured stable code with a safe English fallback; the web API client preserves that code and localizes it through the active zh/en catalog. Responses SSE errors retain the same stable codes, and Docker runtime choices expose message_code for localized UI text. Added API-client and bilingual localization coverage.

internal/llm/service.go
```diff
@@ -1001,6 +1068,10 @@ func writeChatCompletionStreamAsResponse(w io.Writer, r io.Reader, fallbackModel
 			}
 			continue
 		}
+		if code := modelprovider.UpstreamErrorCode([]byte(data)); code != "" {
```
xxx7xxxx: UpstreamErrorCode returns an empty string for valid OpenAI chat error frames whose error.code is missing or null. This branch then parses that frame as a chunk with zero choices, and EOF calls complete(), emitting response.completed with empty or partial text instead of reporting the failure. Please detect the error object independently of whether it carries a code, fall back to upstream_unavailable as rewriteResponsesEventStream does, and add a missing/null-code stream regression test.

wanghaojie124: Fixed in 81005266. Upstream error-object detection is now independent of error.code extraction, so missing or null codes fall back to upstream_unavailable. Chat stream conversion and the model validation probe both use the new detection path. Regression coverage includes missing code, null code after partial output, no erroneous response.completed event, and error:null remaining non-error.

xxx7xxxx (approved):

PR #488, opened 2026-08-18
feat(agent): add secure read-only execution mode

xxx7xxxx:
## Summary

- add configurable standard and secure read-only execution modes for Codex workers
- enforce read-only sandbox, approval, tool, environment, Skill, and MCP boundaries
- add localized profile controls, API documentation, regression coverage, and align the connector action assertion with current main

## Screenshots

image

image

image

image


internal/runtime/codex/appserver_manager.go
```diff
@@ -639,16 +679,70 @@ func appServerTurnStartParams(spec SessionSpec, threadID string, prompt string)
...
+		"--disable", "remote_plugin",
+		"--disable", "shell_tool",
+		"--disable", "skill_mcp_dependency_install",
+		"--disable", "unified_exec",
+		"--disable", "workspace_dependencies",
+		"--config", `approval_policy="never"`,
+		"--config", `sandbox_mode="read-only"`,
+		"--config", `web_search="disabled"`,
+		"--config", `shell_environment_policy.inherit="none"`,
+		"--config", `tools.view_image=false`,
```
wanghaojie124: `tools.view_image` is not a recognized configuration field in the bundled Codex 0.144.1 (`app-server --strict-config` rejects it), so this override is ignored in normal startup and the image reader remains available. That breaks the advertised read-only boundary because an agent can still read local image files. Please enforce this with a supported feature/tool gate (and add an integration check against the bundled binary), or narrow the mode's documented guarantee.

internal/runtime/codex/appserver_manager.go
```diff
@@ -639,16 +679,70 @@ func appServerTurnStartParams(spec SessionSpec, threadID string, prompt string)
...
+		}
 		return map[string]any{"decision": "accept"}, nil
 	case "item/fileChange/requestApproval", "applyPatchApproval":
+		if readOnly {
+			return map[string]any{"decision": "decline"}, nil
+		}
 		return map[string]any{"decision": "accept"}, nil
+	case "item/permissions/requestApproval":
+		if readOnly {
+			return map[string]any{"permissions": []any{}, "scope": "turn"}, nil
```
wanghaojie124: The app-server schema defines `PermissionsRequestApprovalResponse.permissions` as a `GrantedPermissionProfile` object, but this returns an array. If a permission request reaches this path, the response cannot deserialize and the turn will fail instead of receiving an empty grant. Return an empty permission-profile object (and cover this method in the approval-handler test).

internal/runtime/codex/runtime_config.go
```diff
@@ -293,28 +417,57 @@ func configureCodexHomeConfigWithWorkspaceForPlatform(
...
 	content = stripManagedBlock(content, csgclawUserInputBeginMarker, csgclawUserInputEndMarker)
 	content = stripManagedBlock(content, csgclawMemoryFeatureBeginMarker, csgclawMemoryFeatureEndMarker)
 	content = stripManagedBlock(content, csgclawMemoryConfigBeginMarker, csgclawMemoryConfigEndMarker)
 	content = stripManagedBlock(content, csgclawMCPBeginMarker, csgclawMCPEndMarker)
-	if mcpServers != nil {
+	if mcpServers != nil || executionMode == ExecutionModeReadOnly {
 		content = stripTableBlocks(content, mcpServersTableHeaderRe)
 	}
+	if executionMode == ExecutionModeReadOnly && strings.TrimSpace(workspaceDir) != "" {
+		content = stripTableBlocks(content, projectTrustTableHeaderRe(workspaceDir))
```
wanghaojie124: Entering read-only mode permanently removes the existing project table before replacing it with the managed `untrusted` table. On the next switch back to standard mode, the managed block is stripped, but the original table (for example `trust_level = "trusted"` and any other project settings) cannot be restored. As a result, a read-only → standard round trip leaves the workspace untrusted and project-local config disabled. Please preserve the previous project table or enforce untrusted only as a runtime override, and add a round-trip regression test.

internal/runtime/codex/appserver_manager.go
xxx7xxxx: Fixed in 5bb2c692. The unsupported `tools.view_image` override was removed. Read-only mode now uses a supported permission profile that denies `:root`, and the bundled Codex integration test verifies local image reads are rejected.

internal/runtime/codex/appserver_manager.go
xxx7xxxx: Fixed in 5bb2c692. The denial response now returns an empty permission-profile object with turn scope, and the approval-handler regression test covers the response shape.

internal/runtime/codex/runtime_config.go
xxx7xxxx: Fixed in c4a3d2f9. Read-only mode no longer removes or rewrites the existing project table. It applies the current workspace `untrusted` setting only as an app-server startup override, so switching back to standard mode restores normal project-config behavior without losing the original trust entry. Added a read-only to standard round-trip regression test and verified the running Worker transition end to end.

wanghaojie124:
LGTM

wanghaojie124:
LGTM

PR #490, opened 2026-08-18
fix: improve runtime status and billing error handling

wanghaojie124:
- Added detailed Docker container status display, marking agents offline whenever the container is not running.
- Standardized AI Gateway HTTP 402 handling, distinguishing insufficient balance from other payment issues and hiding raw runtime errors in chat.
- Added environment-aware billing links for OpenCSG balance errors, supporting staging, production, and custom login sites.


web/app/src/components/business/MessageContent/MessageContent.tsx
```diff
@@ -119,6 +121,37 @@ export function MessageContent({
   );
 }
 
+function localizeRuntimeError(content: string | null | undefined, t: MessageContentProps["t"]): string {
+  const value = String(content ?? "");
+  if (!t) {
+    return value;
+  }
+  const normalized = value.trim().toLowerCase();
+  if (!normalized.startsWith("runtime error:") && !normalized.includes("unexpected status 402")) {
```
xxx7xxxx: `MessageContent` is also used for human-authored room messages, so matching `unexpected status 402` anywhere in display text changes ordinary messages that quote or discuss this error. It also lets any such message supply a `billing_url` that is rendered as a product-labeled recharge link. Please gate this on trusted runtime-error metadata or at least an actual runtime-error message type instead of parsing every message body.

internal/api/handler.go
```diff
@@ -2972,6 +2972,9 @@ func (h *Handler) backfillAgentLocalUser(resp *agentResponse) {
 
 func (h *Handler) presentAgentResponse(item agent.Agent) agentResponse {
 	resp := presentAgent(item)
+	if resp.Runtime.SandboxEnabled && h != nil && h.svc != nil {
```
xxx7xxxx, line 2996: This decorator is only applied by the list/get/create presenters. The recreate, upgrade, start, stop, and apply-bindings handlers still return `presentAgent(...)` directly, so those mutation responses omit `runtime.sandbox_provider`. The web client usually refetches, but if that detail fetch fails it merges the sparse mutation response after the list refresh and loses the Docker status detail. Please route every agent response through this presenter and cover at least one mutation response.

web/app/src/components/business/MessageContent/MessageContent.tsx
wanghaojie124: Fixed in `bb2c9b9d`. Runtime-error localization now requires trusted server-authored `csgclaw.runtime_error` metadata. The Codex bridge sets this flag only for prompt errors generated by `TurnRenderer`, so human-authored messages and ordinary agent final responses can no longer trigger localization or inject a product-labeled billing link. Added bridge/renderer and `MessageContent` regression coverage.

internal/api/handler.go
wanghaojie124: Fixed in `bb2c9b9d`. The recreate, upgrade, start, stop, and apply-bindings handlers now return `h.presentAgentResponse(...)`, preserving `runtime.sandbox_provider` on mutation responses. Added a start-handler regression assertion for the provider field.

internal/api/handler.go
```diff
@@ -1564,19 +1564,28 @@ func (h *Handler) handleCreateAgentWorker(w http.ResponseWriter, r *http.Request
 	createReq.HubService = hubSvc
 	created, err := h.svc.Create(r.Context(), createReq)
 	if err != nil {
-		writeAgentOperationError(w, err, http.StatusBadRequest)
+		writeAgentOperationErrorWithBillingURL(w, err, http.StatusBadRequest, llm.OpenCSGBillingURL(createReq.Spec.AgentProfile))
```
xxx7xxxx, line 1579: This only adds `billing_url` to create failures. The same profile modal uses `PATCH /api/v1/agents/{id}` in edit mode, but that handler still calls `writeAgentOperationError`, even though `Service.Update` can return the same 402 from `validateRuntimeConfig`; the Agent page save path likewise only keeps the text error. Existing OpenCSG profiles therefore cannot show the new recharge action. Please propagate the billing URL through the update/profile endpoints and wire it into the Agent page error state as well.

wanghaojie124: Fixed in `bf75b0c5`. Both `PATCH /api/v1/agents/{id}` and `PUT /api/v1/agents/{id}/profile` now derive the billing URL from the effective profile after applying field-mask, catalog-selector, and inherited provider-reference rules. The Agent page also preserves the sanitized billing URL in its save-error state, renders the recharge action, and clears stale billing URLs when a non-billing error replaces it. Added backend profile-resolution tests plus controller and UI regression coverage.

xxx7xxxx (approved):

PR #498, opened 2026-08-18
feat(template): improve community publishing workflow

wanghaojie124:
- manage the official template registry from the active OpenCSG environment
- preserve published templates and open their review state when deployment is pending or rejected
- stop retrying deployment after a definitive sensitive-content review failure
- surface duplicate names, sensitive content, and review status with actionable UI feedback


web/app/src/hooks/workspace/useAgentController.ts
```diff
@@ -1735,14 +1741,28 @@ export function useAgentController({
...
-          : target !== "local" && isHubTemplateNameConflict(err)
-            ? t("resourcesPublishCommunityNameExists")
-            : target === "local" && (err as ApiError | null)?.status === 409
-              ? t("agentPublishLocalNameExists")
-              : errorMessage(err, t("agentActionFailed"));
+      const deploySensitiveCheckFailed = isHubTemplateDeploySensitiveCheckError(err);
+      const deployReviewPending = isHubTemplateDeployReviewPendingError(err);
+      const publishedTemplateID = String((err as ApiError | null)?.publishedTemplateId ?? "").trim();
+      if (target === "official_deploy" && publishedTemplateID && (deploySensitiveCheckFailed || deployReviewPending)) {
+        await refreshHubTemplates();
```
xxx7xxxx, line 1750: This branch assumes `refreshHubTemplates()` makes `publishedTemplateID` selectable, but `AGENT-ERR-22` is returned precisely while that repo is still absent from the agent-template catalog. `useWorkspaceHubSelection` resets any selected ID not present in `resourcesTemplates` to the first item, so this returns `true` and navigates while the UI immediately shows a different template and no pending-review state. Please preserve or seed the published template until the catalog catches up, or allow direct-detail selection for IDs not yet in the list. The Hub publish controller has the same assumption.

internal/api/handler.go
```diff
@@ -1221,12 +1221,30 @@ func (h *Handler) publishUpdatedAgentUser(updated agent.Agent) {
...
+		http.Error(w, "agent service is not configured", http.StatusServiceUnavailable)
+		return
+	}
+	if r.Method == http.MethodGet {
+		if err := h.svc.Reload(); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+	}
+	selected, ok := h.svc.Agent(selector)
```
xxx7xxxx: `Agent()` and `AgentByName()` hydrate runtime status, so using them only to resolve this selector adds an extra sandbox probe before `handleAgentProfile` performs its own profile work. In the added test, GET emitted two failed probes and took 4.25 seconds, while PUT emitted three and took 6.03 seconds when the sandbox was unavailable. Please resolve the canonical ID through a lightweight snapshot or metadata lookup so profile reads and updates do not block on runtime health.

web/app/src/hooks/workspace/useAgentController.ts
wanghaojie124: Fixed in 790038b7. `useWorkspaceHubSelection` now preserves a non-empty direct-detail template ID even when it is not yet present in the catalog, and it no longer falls back to rendering the first catalog template for that pending ID. This covers both the agent publish and Hub publish flows. Added regression coverage for listed and not-yet-listed selections.

internal/api/handler.go
wanghaojie124: Fixed in 790038b7. Profile selector resolution now uses the lightweight `ResolveAgentID` snapshot lookup, so resolving an ID or display name does not hydrate runtime status or probe the sandbox before the profile operation. The name-resolution GET/PUT regression test remains covered, and the targeted API test passes.

web/app/src/hooks/workspace/useWorkspaceHubSelection.ts
```diff
@@ -78,7 +82,10 @@ export function useWorkspaceHubSelection({
     );
   }, [officialSkillsQuery.data]);
   const selectedHubTemplate = useMemo(
-    () => resourcesTemplates.find((item) => item.id === selectedHubTemplateId) || resourcesTemplates[0] || null,
+    () =>
```
xxx7xxxx, line 92: Preserving the missing ID fixes the wrong-template fallback, but the pending review state is still unavailable. `AGENT-ERR-22` means the catalog refresh has no entry, while the direct detail path uses `RemoteStore.Get`, whose `remoteCodeRepository.Metadata` is excluded from JSON and only populated by the catalog-list merge. The selected detail therefore has no `sensitive_check` metadata, and `publishAgentPage` returns success without showing the pending message. Please seed a pending banner or metadata from the structured error, or make the detail API return review status, and cover the complete controller-to-detail flow.

internal/llm/service.go
```diff
@@ -600,6 +612,74 @@ func compactUpstreamErrorStreamFromBody(resp *http.Response, body []byte, billin
...
+		if value := strings.TrimSpace(header.Get(key)); value != "" {
+			return value
+		}
+	}
+	return ""
+}
+
+func upstreamErrorBodyForLog(body []byte) string {
+	truncated := len(body) > maxUpstreamErrorLogBytes
+	if truncated {
```
xxx7xxxx, line 633: Truncating before structured redaction bypasses the secret-key scrub. For bodies over 4 KB, JSON parsing is skipped, and the fallback regex does not match normal quoted JSON separators such as `"api_key":"sk-..."`. I reproduced that synthetic key remaining in the returned log value. Since upstream error bodies can echo request configuration, this can write credentials to logs. Please redact parsed JSON before truncating, or harden the fallback for quoted JSON, and add a test that combines truncation with a secret.

web/app/src/shared/i18n/index.ts
```diff
@@ -79,6 +79,55 @@ export function localizeError(raw: unknown, t: TranslateFn): string {
...
+  const match = value.match(/unexpected status\s+(\d{3})\b/i);
+  const status = Number(match?.[1] ?? 0);
+  if (status === 400 || status === 422) return t("errors.invalid_request_error");
+  if (status === 401) return t("errors.authentication_error");
+  if (status === 402) return t("errors.payment_required");
+  if (status === 403) return t("errors.forbidden");
+  if (status === 404) return t("errors.not_found");
+  if (status === 429) return t("errors.rate_limit_exceeded");
+  if (status >= 500 && status <= 599) return t("errors.upstream_unavailable");
+  return t("errors.upstream_unavailable");
```
xxx7xxxx: This fallback labels every unrecognized trusted runtime error as a temporarily unavailable model service, even when the `unexpected status` regex did not match. The same `runtime_error` metadata is used for slash-command parsing, session resolution, conversation reset, sandbox, and other prompt failures, so local errors such as `Runtime error: sandbox process exited unexpectedly` now display and notify the wrong diagnosis. Please only apply the HTTP mapping when a status was parsed; otherwise preserve the actual error or use a generic runtime-error message.

web/app/src/hooks/workspace/useWorkspaceHubSelection.ts
wanghaojie124: Fixed in 2a81643f. Both publish controllers now seed or update the Hub templates query cache with the structured review result, Pending or Fail, before selecting the published template. mergeHubTemplateDetail preserves that seeded sensitive_check metadata when the direct detail response omits it, so the detail pane can render the pending or failure state while the catalog catches up. Added model coverage for creating the pending summary, merging it into detail, and preserving existing catalog fields.

internal/llm/service.go
wanghaojie124: Fixed in 2a81643f. The fallback redactor now handles quoted JSON keys and quoted values after truncation, including api_key, cookie, authorization, token, secret, and password fields. Added a regression case with a body over 4 KB containing both an API key and cookie, asserting neither secret survives and the truncation marker remains.

web/app/src/shared/i18n/index.ts
wanghaojie124: Fixed in 2a81643f. Runtime error presentation is now produced centrally by runtimebridge. HTTP-specific mappings are applied only when an unexpected status code is actually parsed; all other prompt or runtime failures use the generic internal_error presentation instead of being labeled as upstream unavailability. Added explicit coverage for a local sandbox-process failure and for recognized versus unrecognized HTTP statuses.

xxx7xxxx (approved):

PR #501, opened 2026-08-19
fix(team): constrain team member avatar rendering

youngbeom-shin:
## Summary
- fix Team detail member avatar styles to target global avatar classes inside the CSS module
- add overflow clipping and block image rendering so member avatars stay constrained to the 40x40 tile on initial load
- prevent the intermittent oversized or incorrect-looking avatar rendering on the team members panel


youngbeom-shin:
before：
截屏2026-08-19 14 54 05


after：
截屏2026-08-19 14 53 06

@jialudev please help review

jialudev (approved):
LGTM

PR #506, opened 2026-08-20
feat(feishu): route direct channel through agent engine

GatewayJ:
## Summary

This PR refactors the Feishu channel from the legacy `codexbridge` execution path into a binding-owned direct AgentEngine channel.

The new Feishu data plane is split into dedicated packages for binding lifecycle, transport, ingress, context hydration, file preparation, execution, presentation, delivery, and process-local state. Hosted Feishu workers now route incoming Feishu messages directly into AgentEngine Conversations, while Feishu app/bot/team/room management remains in the existing control-plane service.

### Channel lifecycle

- Add a dedicated `FeishuBindingManager` started from `serve`.
- Reconcile Feishu channel workers from AgentEngine agent bindings and participant credentials.
- Route hosted Feishu workers through the new direct AgentEngine pipeline.
- Remove the legacy Feishu `codexbridge` execution path without adding fallback behavior.

### Ingress and conversation context

- Add a bounded concurrent Intake pipeline for Feishu events.
- Route WebSocket events and local MessageBus handoff through the same intake path.
- Improve bot self-message filtering, exact mention matching, event deduplication, freshness checks, and per-scope watermarks.
- Support Feishu message events, document comments, and compatible card actions in the transport event model.
- Hydrate quoted messages when available and include structured channel metadata in AgentEngine prompt context.
- Clarify conversation scoping:
  - true Feishu threads create separate conversations;
  - root/parent messages are treated as quote relationships;
  - document comments use their own comment scope.

### File handling

- Support inbound Feishu images/files as AgentEngine input files.
- Download Feishu resources into a binding-scoped staging area.
- Enforce attachment count, file size, message total size, binding quota, MIME detection, SHA256 calculation, and regular-file validation.
- Clean temporary attachment files after runs and clean stale Feishu attachment staging files on startup.
- Align with the refactored AgentEngine file API by passing Agent-scoped file IDs instead of raw source-path metadata.

### Execution and delivery

- Run Feishu turns through AgentEngine Conversations using `AdmissionSupersede` and create-or-resume continuation semantics.
- Use Engine Reset for `/new` and Engine Cancel for cancellation.
- Standardize hosted Feishu replies on Markdown.
- Stream text, thought, and tool status updates with throttling.
- Split oversized Markdown payloads into multiple Feishu messages.
- Track process-local delivery state, remote message IDs, update dependencies, reaction operations, and terminal snapshots.
- Retry retryable Feishu OpenAPI operations with stable UUIDs and bounded backoff.

### Observability

- Add structured Feishu channel logs across worker lifecycle, ingress decisions, context hydration, file preparation, AgentEngine execution, delivery dispatch, and Feishu API operations.
- Include binding, agent, app, event, turn, delivery, and API fields to make production troubleshooting easier.

## Impact

- Hosted Feishu workers now use the new direct AgentEngine channel path.
- Legacy Feishu execution through `codexbridge` is removed.
- Delivery and turn state are process-local best effort; this PR does not add WAL, checkpointing, or restart recovery.
- Hosted output is fixed to Markdown. CardKit-related code remains in the package but is not the active presentation mode.
- Outbound Feishu image/file upload is not implemented in this PR.
- `resource_link` and interactive user-input rendering are currently ignored by the hosted Feishu presentation path.

## Validation

- `go test ./internal/channel/feishu/...`
- `go test ./cli/serve -run Feishu`
- `git diff --check`


jialudev:
LGTM

xxx7xxxx:
Thanks for landing this. I re-reviewed the merged Feishu path on current `main` (`b51584e1`) and found two lifecycle/concurrency issues that are worth a focused follow-up PR:

- **Bound active Feishu Runs across conversations.** `internal/channel/feishu/execution/runner.go:142` starts a long-lived goroutine for every accepted message and returns immediately, so the 32-item Intake queue and 8 Intake handlers do not bound Engine work. `AdmissionSupersede` only limits the same conversation; distinct chats/topics can create unbounded active Turns, goroutines, Turn records, and delivery intents. A blocking-Engine reproduction on current `main` reached 200 simultaneous Runs from 200 submissions. The new built-in CSGClaw ingress worker now caps active conversation scopes at 8, which looks like a useful pattern for the Feishu scheduler as long as `/new` and cancel remain able to bypass a saturated work queue.

- **Make the Feishu WebSocket lifecycle joinable and observable after readiness.** `internal/channel/feishu/transport/ingress_lark.go:83` launches SDK `Client.Start`, but the pinned SDK v3.9.7 blocks forever in `select {}` after connecting and its ping loop ignores cancellation. Binding restarts therefore leak the old client goroutines. More importantly, a terminal reconnect failure only stops the SDK receive loop; CSGClaw still considers the binding active and silently stops receiving events. v3.10.0 remains affected. Upstream tracks this in https://github.com/larksuite/oapi-sdk-go/issues/231 and https://github.com/larksuite/oapi-sdk-go/pull/232. The follow-up should both use a genuinely stoppable SDK lifecycle and supervise post-ready completion/errors so the binding can be restarted or marked unhealthy.

Could you send a focused follow-up PR covering these two cases with blocking-Run and Start/Disconnect/reconnect regression tests?

PR #507, opened 2026-08-20
feat(channel): add built-in IM adapter core

jialudev:
- Route eligible host Codex participants through a binding-scoped built-in IM adapter backed by Agent Engine while preserving existing direct, mention, room, thread, and reply behavior.
- Render final answers, thoughts, tool activity, permissions, and structured user input through the existing message and activity APIs with stable IDs and legacy-compatible metadata.
- Materialize source-scoped attachments, serialize turns with wait admission, prevent stale turn updates, and release runtime work state across success, failure, and cancellation.

### DIFF
--- a/cli/serve/serve.go
+++ b/cli/serve/serve.go
@@ -34,6 +34,12 @@ import (
 	"csgclaw/internal/app/channelwiring"
 	"csgclaw/internal/app/runtimewiring"
 	csgclawchannel "csgclaw/internal/channel/csgclaw"
+	"csgclaw/internal/channel/csgclaw/binding"
+	"csgclaw/internal/channel/csgclaw/delivery"
+	"csgclaw/internal/channel/csgclaw/execution"
+	"csgclaw/internal/channel/csgclaw/files"
+	csgclawinteraction "csgclaw/internal/channel/csgclaw/interaction"
+	csgclawsource "csgclaw/internal/channel/csgclaw/source"
 	"csgclaw/internal/channel/feishu"
 	feishubinding "csgclaw/internal/channel/feishu/binding"
 	"csgclaw/internal/channel/feishu/participantprovider"
@@ -51,6 +57,7 @@ import (
 	"csgclaw/internal/modelprovider"
 	internalonboard "csgclaw/internal/onboard"
 	"csgclaw/internal/participant"
+	agentruntime "csgclaw/internal/runtime"
 	runtimecodex "csgclaw/internal/runtime/codex"
 	"csgclaw/internal/runtimecatalog"
 	"csgclaw/internal/sandboxproviders"
@@ -104,6 +111,7 @@ var (
 	}
 	NewCodexBridgeManager   = newCodexBridgeManager
 	NewFeishuBindingManager = newFeishuBindingManager
+	NewCSGClawAdapterSource = newCSGClawAdapterSource
 	OpenBrowser             = openBrowser
 	WaitForHealthy          = waitForHealthy
 	stopServerProcessByPID  = stopServerProcess
@@ -113,6 +121,11 @@ type serveCmd struct{}
 type stopCmd struct{}
 type internalServeCmd struct{}
 type desktopServeCmd struct{}
+
+type csgclawAdapterSource interface {
+	Start(context.Context) error
+	Close()
+}
 
 const cliproxyAutoLoginEnv = "CSGCLAW_CLIPROXY_AUTO_LOGIN"
 
@@ -605,6 +618,7 @@ func startServerWithConfigPath(ctx context.Context, run *command.Context, cfg co
 	if err != nil {
 		return err
 	}
+	participantBridge := im.NewParticipantBridge(cfg.Server.AccessToken)
 	workBus := worklease.NewBus()
 	workControlBus := worklease.NewControlBus()
 	turnControlDispatcher := worklease.NewTurnControlDispatcher(workControlBus)
@@ -646,6 +660,16 @@ func startServerWithConfigPath(ctx context.Context, run *command.Context, cfg co
 		if err := feishuBindingMgr.Start(ctx); err != nil {
 			return err
 		}
+	}
+	imAdapterSource, err := NewCSGClawAdapterSource(conversationEngine, svc, imSvc, imBus, participantSvc, workRegistry, participantBridge)
+	if err != nil {
+		return err
+	}
+	if imAdapterSource != nil {
+		if err := imAdapterSource.Start(ctx); err != nil {
+			return err
+		}
+		defer imAdapterSource.Close()
 	}
 	llmSvc, err := NewLLMService(cfg, svc)
 	if err != nil {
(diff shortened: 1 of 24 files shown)


GatewayJ:
High: [ingress.go (line 91)](/tmp/csgclaw-pr507-review-0824/internal/channel/csgclaw/ingress/ingress.go:91) /new 只取消当前 active turn，但不会清掉同一 conversation 里已排队的旧消息。场景：m1 正在跑、m2 已排队、用户发 /new，m1 被取消后调度器仍会先跑 m2，再执行 reset，导致 reset 语义被旧消息穿透。建议 reset 入队时丢弃/跳过该 conversation 的旧 pending，或给 conversation 做 generation fence。
Medium: [source.go (line 311)](/tmp/csgclaw-pr507-review-0824/internal/channel/csgclaw/source/source.go:311) worker 拒绝事件时调用 ParticipantBridge.Requeue，但 [ParticipantBridge (line 127)](/tmp/csgclaw-pr507-review-0824/internal/im/participant_bridge.go:127) 只在新的 Subscribe 时灌 pending，当前订阅保持活跃时不会自动重试。这个“回退”会把消息搁置到 worker/source 重建，队列满或临时 binding 不可用时用户消息可能长期不处理。建议改成明确 drop+ack+告警，或在 Source 内做带退避的 retry。
Medium: [ingress.go (line 197)](/tmp/csgclaw-pr507-review-0824/internal/channel/csgclaw/ingress/ingress.go:197) adapter.Handle 的错误被吞掉，而 [source.go (line 316)](/tmp/csgclaw-pr507-review-0824/internal/channel/csgclaw/source/source.go:316) 已经在 submit 成功后 ack。若 final 渲染/IM 持久化失败，用户看不到回答，也没有日志或补偿。建议至少记录 binding/message/turn 维度的错误；不要重跑 Agent，但要让失败可诊断。

jialudev:
这三个问题都成立，已在最新提交中修复：

- `/new` 现在通过 conversation generation fence 使同一会话中更早的 pending 消息失效，同时仍会取消 active turn；reset 后的新消息进入新 generation。
- Source 在订阅仍活跃时会对临时 binding 不可用或 worker 拒绝执行做指数退避重试；只有订阅/Source 停止时才 Requeue 给下次订阅，避免消息长期搁置。
- ingress 不会重跑 Agent，但会记录包含 binding、participant、message、conversation 和 turn 维度的 `adapter.Handle` 错误，便于诊断渲染或 IM 持久化失败。

另外，该分支已 rebase 到最新 `upstream/main`，PR base 已改为 `main`。

jialudev:
@GatewayJ 感谢 review。这三个问题都成立，已经在 e44dd12c 中分别通过 conversation generation fence、活跃订阅内指数退避重试，以及带 binding/message/conversation/turn 上下文的错误日志修复。分支也已 rebase 到最新 upstream/main，PR base 已改为 main。

GatewayJ:
lgtm

cli/serve/serve.go
```diff
@@ -647,6 +661,16 @@ func startServerWithConfigPath(ctx context.Context, run *command.Context, cfg co
 			return err
 		}
 	}
+	imAdapterSource, err := NewCSGClawAdapterSource(conversationEngine, svc, imSvc, imBus, participantSvc, workRegistry, participantBridge)
```
xxx7xxxx, line 664: Starting the new in-process source here is not an atomic switch: newCodexBridgeManager still configures CSGClawClient, and that manager is started in OnReady. Both paths subscribe to the same ParticipantBridge, which fans each event out to every subscriber, so every eligible host Codex message can run twice and emit duplicate turns and replies. Please remove the legacy CSGClaw bridge and lifecycle observer before starting this source, and add an end-to-end assertion that only one consumer handles a message.

internal/channel/csgclaw/source/source.go
```diff
@@ -0,0 +1,471 @@
...
+func (s *Source) submitWithRetry(ctx context.Context, bindingID channel.BindingID, event im.ParticipantEvent) bool {
+	delay := s.retryInitial
+	if delay <= 0 {
+		delay = defaultRetryInitial
+	}
+	maxDelay := s.retryMax
+	if maxDelay < delay {
+		maxDelay = defaultRetryMax
+	}
+	for {
```
xxx7xxxx, line 335: This retry loop stops the only goroutine draining the 16-slot ParticipantBridge subscription while the worker is unavailable or full. Once that channel fills, later events move to ParticipantBridge.pending, which is drained only by Subscribe; when this retry eventually succeeds, the existing subscription stays open, so those messages can remain stranded, and the 64-item pending cap can drop older ones. Please keep draining into a bounded source queue or explicitly resubscribe and drain pending after backpressure clears.

internal/channel/csgclaw/interaction/interaction.go
```diff
@@ -0,0 +1,258 @@
...
+			IsOther:  question.IsOther,
+			IsSecret: question.IsSecret,
+		})
+	}
+	var autoResolve time.Duration
+	if args.AutoResolutionMS != nil {
+		autoResolve = time.Duration(*args.AutoResolutionMS) * time.Millisecond
+	}
+	snapshot, err := c.broker.CreateDetached(runtimecodex.PendingUserInputRequest{
+		Execution: activity.ExecutionRef{
```
xxx7xxxx, line 124: This detached request is still registered without Execution.RuntimeID or Execution.SessionID. /new resets the runtime conversation through UserInputBroker.CancelSession(runtimeID, sessionID), so these requests do not match that cancellation and remain pending and actionable after the conversation has been cleared, indefinitely when no auto-resolution is configured. Please associate the request with the runtime/session cancellation identity, or explicitly interrupt Coordinator requests for this conversation during reset.

xxx7xxxx:
Oops, comments and merge happened at the same time. You could fix them in the next pr if they are valid @jialudev

jialudev:
> Oops, comments and merge happened at the same time. You could fix them in the next pr if they are valid @jialudev

Thanks for the review. I’ll handle these in a follow-up PR.

internal/channel/csgclaw/interaction/interaction.go
jialudev: Thanks, fixed in #521. Before creating a detached request, the coordinator now resolves the active Agent RuntimeID and Codex Engine SessionID and stores both in Execution. The existing /new ResetConversation path can therefore interrupt the request through CancelSession; the regression test verifies cancellation only matches the real runtime/session identity.

cli/serve/serve.go
jialudev: Thanks, fixed in #521. The serve path no longer constructs or starts the legacy codexBridgeMgr, and CSGClaw binding refresh now reconciles the in-process adapter source. The regression test asserts one ParticipantBridge subscriber and one handling of a published message.

internal/channel/csgclaw/source/source.go
jialudev: Thanks, fixed in #521. ParticipantBridge now refills pending events after each successful acknowledgement, so once worker retry backpressure clears, the active subscription continues draining queued events without requiring resubscription. The regression test holds the worker unavailable, publishes beyond the 16-slot subscription buffer, then verifies every message is submitted after recovery.

PR #510, opened 2026-08-21
fix(im): prevent duplicate message sends

youngbeom-shin:
## Summary
- add a synchronous per-conversation send lock to block rapid repeated clicks
- attach a stable client message ID to normal messages and thread replies
- deduplicate repeated message requests server-side and suppress duplicate events

### DIFF
--- a/internal/api/handler.go
+++ b/internal/api/handler.go
@@ -744,14 +744,15 @@ func bootstrapRuntimeKind(runtime string) string {
 }
 
 type createMessageRequest struct {
-	RoomID      string                       `json:"room_id"`
-	SenderID    string                       `json:"sender_id"`
-	Content     string                       `json:"content"`
-	Locale      string                       `json:"locale,omitempty"`
-	MentionID   string                       `json:"mention_id,omitempty"`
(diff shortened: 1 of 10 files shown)


youngbeom-shin:
Please review this PR. Thanks!

jialudev:
There is a failure-recovery issue in `CreateMessageOnce` that should be addressed before merging.

The message is appended to `room.Messages` before `saveLocked()` is called. If persistence fails, the method returns an error but leaves the message and its `client_message_id` in memory. A retry with the same client message ID will then match that in-memory message and return it with `created=false`. As a result, the handler responds with 201 without retrying persistence, publishing `message.created` / thread updates, or handling the team-room command. The message can also disappear after a restart because it was never persisted.

Please roll back the in-memory message, thread, and attachment-related mutations when saving fails, or otherwise ensure that only successfully persisted messages participate in idempotency lookup. A focused regression test covering “first save fails, then the same client message ID is retried” would also help protect this behavior.

jialudev (approved):
LGTM

PR #512, opened 2026-08-24
fix(web): align human avatar and identity layout

youngbeom-shin:
## Summary
- show the same fallback initial avatar in the human list and detail header
- keep fallback avatars editable with the existing avatar picker
- make role and user ID cards adapt to the available container width
- wrap long user IDs safely inside their card


youngbeom-shin:
before:
截屏2026-08-24 13 31 38

after:
截屏2026-08-24 13 30 35

jialudev (approved):
LGTM

PR #515, opened 2026-08-25
feat(feishu): support file transmission

GatewayJ:
## Summary

- deliver successful Agent Engine output artifacts through Feishu image or file messages using Agent-scoped authoritative file snapshots
- normalize both direct and locale-wrapped Feishu rich-text bodies so inbound images retain their `image_key`, download through the message-resource API, and reach Codex as image inputs
- split media upload from message send and reuse uploaded keys when send retries, avoiding duplicate uploads
- enforce Feishu and per-turn size/count limits with user-visible warnings
- reject zero-byte output artifacts before queueing and at both raw upload boundaries, using the same warning path instead of a permanent remote failure
- accept Feishu processing-reaction lifecycle events without treating echoed presentation state as an ingress failure
- tell Codex to publish generated files immediately through the active channel using both the dynamic tool description and shared Runtime managed instructions, without searching for CLI, curl, HTTP, or channel-specific upload alternatives

## Validation

- `go test -count=1 ./internal/agent ./internal/channel/feishu/... ./internal/agentengine/... ./internal/runtime/codex`
- `go test -race -count=1 ./internal/agent ./internal/channel/feishu/transport ./internal/channel/feishu/ingress ./internal/channel/feishu/execution ./internal/runtime/codex ./internal/agentengine`
- `go vet ./internal/agent ./internal/channel/feishu/... ./internal/agentengine/... ./internal/runtime/codex`
- `git diff --check`

## Notes

- supported non-empty images up to 10 MiB use Feishu image messages; other non-empty artifacts up to 30 MiB use generic file messages
- Engine file metadata is the single source of truth; delivery does not fall back to intent metadata or cross-Agent lookup
- rich-text normalization accepts the two Feishu wire shapes used here (direct body and locale wrapper) without recursively guessing arbitrary payloads
- processing reactions are consumed only at the Feishu transport boundary; they do not enter Agent input
- shared output-delivery guidance applies to Manager and Worker Codex Runtimes when `csgclaw_publish_file` is available; Manager connector and historical-attachment rules remain Manager-only
- duplicate user requests keep the existing latest-turn-wins behavior
- this completes the Feishu channel integration for the Agent Engine file capability introduced by #509

### DIFF
--- a/internal/channel/feishu/execution/runner.go
+++ b/internal/channel/feishu/execution/runner.go
@@ -14,10 +14,15 @@ import (
 	feishuctx "csgclaw/internal/channel/feishu/context"
 	"csgclaw/internal/channel/feishu/presentation"
 	feishustate "csgclaw/internal/channel/feishu/state"
+	"csgclaw/internal/channel/feishu/transport"
 	"csgclaw/internal/slashcommand"
 )
 
-const processingPinEmoji = "Pin"
+const (
+	processingPinEmoji       = "Pin"
+	maxFeishuOutputFileCount = 8
+	maxFeishuOutputFileTotal = int64(50 << 20)
+)
 
 type FilePreparer interface {
 	Prepare(context.Context, channeltypes.InboundMessage) ([]agentengine.InputPart, func(), error)
@@ -376,6 +381,9 @@ func (r *Runner) finalize(message channeltypes.InboundMessage, result agentengin
 			terminal = presentation.Terminal(mode, presentationResult(result))
 		}
 		intents = append(intents, r.presentationUpdateIntents(message, finalSequence, true, terminal)...)
+		if status == channeltypes.TurnSucceeded {
+			intents = append(intents, r.fileDeliveryIntents(message, result.Files, finalSequence+1)...)
+		}
 		if cleanup, ok := r.reactionCleanupIntent(message); ok {
 			intents = append(intents, cleanup)
 		}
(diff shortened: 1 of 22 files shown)


internal/channel/feishu/execution/runner.go
```diff
@@ -462,6 +470,68 @@ func (r *Runner) textIntent(message channeltypes.InboundMessage, id string, sequ
...
+		intents = append(intents, warning)
+	}
+	return intents
+}
+
+func outputFileRejection(file agentengine.OutputFile, accepted int, acceptedBytes int64) string {
+	if strings.TrimSpace(file.ID) == "" {
+		return "Engine file ID is empty"
+	}
+	if file.SizeBytes < 0 {
```
xxx7xxxx: (nit): Feishu's image and file upload endpoints reject zero-byte bodies, but this guard accepts `SizeBytes == 0` and queues the artifact. Empty generated artifacts are valid Engine output, so they reach the API, fail permanently, and unlike files rejected here no user-visible warning is queued. Please reject `<= 0` here and at the upload boundary, with a regression test, so zero-byte outputs use the existing warning path.

xxx7xxxx (approved):

internal/channel/feishu/execution/runner.go
GatewayJ: Fixed in 5d00bdde.

- Feishu execution now rejects `SizeBytes <= 0`, so empty Engine outputs use the existing user-visible warning path instead of queueing a media delivery.
- Both raw image and file upload boundaries require a positive size, and `SupportsImageUpload` is consistent with that contract.
- Added regression coverage for the warning intent and for preventing both OpenAPI upload calls.

Validation: focused/full tests, race tests, vet, and `git diff --check` all pass.

PR #517, opened 2026-08-25
fix(template): structure publishing and deployment errors

wanghaojie124:
- preserve upstream Hub error codes and return a consistent structured API error contract
- distinguish pending sensitive-content reviews from failed reviews after deployment retries
- localize publishing, review, conflict, account, and resource errors through shared error-code messages
- keep published template ids on partial deployment failures and retain server review details
- teach CLI clients to read nested structured errors and preserve nonstandard upstream diagnostics
- add backend and frontend regression coverage for error classification, localization, and compatibility

### DIFF
--- a/web/app/src/hooks/workspace/useAgentController.ts
+++ b/web/app/src/hooks/workspace/useAgentController.ts
@@ -1741,16 +1741,18 @@ export function useAgentController({
       const deployReviewPending = errorCode === HubTemplateErrorCodes.reviewPending;
       const publishedTemplateID = String((err as ApiError | null)?.publishedTemplateId ?? "").trim();
       const message = errorMessage(err, t("agentActionFailed"));
-      if (target === "official_deploy" && publishedTemplateID && (deploySensitiveCheckFailed || deployReviewPending)) {
+      if (target === "official_deploy" && publishedTemplateID) {
         await refreshHubTemplates();
-        queryClient.setQueryData<HubTemplate[]>(workspaceQueryKeys.hubTemplates(), (templates) =>
-          upsertHubTemplateReviewState(
-            templates,
-            publishedTemplateID,
-            deployReviewPending ? "Pending" : "Fail",
-            deployReviewPending ? "" : message,
-          ),
-        );
+        if (deploySensitiveCheckFailed || deployReviewPending) {
+          queryClient.setQueryData<HubTemplate[]>(workspaceQueryKeys.hubTemplates(), (templates) =>
+            upsertHubTemplateReviewState(
+              templates,
+              publishedTemplateID,
+              deployReviewPending ? "Pending" : "Fail",
+              deployReviewPending ? "" : message,
+            ),
+          );
+        }
         setSelectedHubTemplateId(publishedTemplateID);
         navigatePane({ type: WorkspacePaneTypes.hub, id: publishedTemplateID, resourceType: "template" }, rooms);
         return true;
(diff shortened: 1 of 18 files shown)


web/app/src/hooks/workspace/useWorkspaceHubController.ts
```diff
@@ -235,21 +237,10 @@ export function useWorkspaceHubController({
...
-              ? t("agentPublishSensitiveInformation")
-              : isHubTemplateAccountEmailMissing(err)
-                ? t("resourcesPublishCommunityEmailRequired")
-                : isHubTemplateNameConflict(err)
-                  ? t("resourcesPublishCommunityNameExists")
-                  : errorMessage(err, t("resourcesPublishCommunityFailed"));
+        const errorCode = hubTemplateErrorCode(err);
+        const deploySensitiveCheckFailed = errorCode === HubTemplateErrorCodes.reviewFailed;
+        const deployReviewPending = errorCode === HubTemplateErrorCodes.reviewPending;
+        const message = errorMessage(err, t("resourcesPublishCommunityFailed"));
```
xxx7xxxx, line 244: This only treats the two review codes as a partial result. `RESOURCE-ERR-1` and the new `template_deploy_failed` response also include `published_template_id`, because the template was already published before deployment failed. They therefore fall through to `return null`, leaving the publish dialog open and presenting an overall failure; the agent-page path has the same issue and returns `false`. Please use the published ID (or all deployment-failure codes) to report partial success and select the published template on both entry points.

wanghaojie124: Fixed in 9639cc38. Both publishing entry points now treat any deployment error carrying `published_template_id` as a partial success, select and open the published template, and only apply the local review-state override for the review-pending/review-failed codes. I also added regression coverage for `RESOURCE-ERR-1` and `template_deploy_failed`; the full frontend suite passes (946 tests).

web/app/src/hooks/workspace/useAgentController.ts
```diff
@@ -1743,36 +1736,27 @@ export function useAgentController({
...
     } catch (err) {
-      const deploySensitiveCheckFailed = isHubTemplateDeploySensitiveCheckError(err);
-      const deployReviewPending = isHubTemplateDeployReviewPendingError(err);
+      const errorCode = hubTemplateErrorCode(err);
+      const deploySensitiveCheckFailed = errorCode === HubTemplateErrorCodes.reviewFailed;
+      const deployReviewPending = errorCode === HubTemplateErrorCodes.reviewPending;
       const publishedTemplateID = String((err as ApiError | null)?.publishedTemplateId ?? "").trim();
-      if (target === "official_deploy" && publishedTemplateID && (deploySensitiveCheckFailed || deployReviewPending)) {
+      const message = errorMessage(err, t("agentActionFailed"));
+      if (target === "official_deploy" && publishedTemplateID) {
```
xxx7xxxx, line 1745: This preserves the published template ID, but it still reports every partial deployment as full success: returning `true` closes the Agent publish dialog, and this branch never surfaces `message` through an error or notice. The review-pending/review-failed cases remain visible through the Hub review state, but `RESOURCE-ERR-1` and `template_deploy_failed` have no corresponding state there, so users are never told that deployment failed. Please propagate a partial result/message like `useWorkspaceHubController`, or show an equivalent persistent notice while still navigating to the published template.

wanghaojie124: Fixed in a8263f08. For non-review partial deployment failures, the Agent publish path now forwards the localized deployment error into the Hub controller before navigating, so the selected published template opens with a persistent Hub mutation error instead of appearing as an unqualified success. Review-pending/review-failed responses continue to use their existing Hub review state. The regression test now verifies that both `RESOURCE-ERR-1` and `template_deploy_failed` propagate the message, and the full frontend suite passes (946 tests).

xxx7xxxx (approved):

PR #519, opened 2026-08-25
fix(desktop): prevent Windows proxy from breaking agent network access

jialudev:
- Prevent Windows system proxy settings from routing all CSGClaw traffic through a local proxy and breaking direct site access.
- Keep the proxy scoped to Codex and Claude Code providers while other services use their default network path.

### DIFF
--- a/desktop/src/main/sidecar/childEnvironment.ts
+++ b/desktop/src/main/sidecar/childEnvironment.ts
@@ -3,7 +3,8 @@ import { userInfo } from "node:os";
 import path from "node:path";
 import { DesktopPlatform } from "../../shared/desktopEnvironment";
 import {
-  resolveSystemProxyEnvironment,
+  CLIPROXY_SYSTEM_PROXY_ENV,
+  resolveSystemCLIProxyEnvironment,
   type SystemProxyResolver,
 } from "./systemProxy";
 
@@ -17,6 +18,14 @@ const SHELL_ENVIRONMENT_KEYS = [
   "DOCKER_CONFIG",
   "DOCKER_CERT_PATH",
   "DOCKER_TLS_VERIFY",
+] as const;
+const STANDARD_PROXY_ENVIRONMENT_KEYS = [
+  "HTTPS_PROXY",
+  "https_proxy",
+  "HTTP_PROXY",
+  "http_proxy",
+  "ALL_PROXY",
+  "all_proxy",
 ] as const;
 const WINDOWS_ENVIRONMENT_SCRIPT = `
 [Console]::OutputEncoding = [Text.UTF8Encoding]::new($false)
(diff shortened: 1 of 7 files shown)


desktop/src/main/sidecar/childEnvironment.ts
```diff
@@ -73,7 +82,22 @@ export async function resolveSidecarEnvironment({
 }: ResolveSidecarEnvironmentOptions): Promise<NodeJS.ProcessEnv> {
   // Restore the environment that a newly opened terminal would normally
   // provide, while keeping packaged helper executables ahead of host tools.
+  const inheritedProxyURL = standardProxyEnvironmentValue(
```
xxx7xxxx: This also reinterprets explicitly inherited `HTTPS_PROXY`/`HTTP_PROXY`/`ALL_PROXY` values as CLIProxy-only settings, although the stated scope is the proxy returned by Electron. The general LLM client intentionally uses Go's default transport so proxy environment variables and `NO_PROXY` are honored; after this rewrite, a Desktop launch in an environment that requires an explicit proxy sends OpenCSG and custom-provider traffic directly and can no longer connect. Please scope only the `session.resolveProxy` result, or otherwise preserve an explicit inherited proxy path for general sidecar traffic.

xxx7xxxx:
After comparing this with #423, the scoping direction makes sense, but the implementation needs one boundary adjustment.

#423 intentionally made a Finder-launched Desktop app honor the OS system proxy by resolving it through Electron and exposing it to the whole sidecar as standard proxy environment variables. This PR changes that contract: the Electron-resolved proxy is passed only to the embedded CLIProxy, so Codex and Claude Code use it while OpenCSG, CSGHub Lite, and custom providers connect directly. That prevents a stale or broken local Windows proxy from taking down all provider traffic, but the implementation affects macOS and Linux as well as Windows.

The resulting behavior is conditional: if the local proxy endpoint is broken, non-CLIProxy providers can recover through direct access, while Codex and Claude Code still fail through that proxy. On networks where direct access is unavailable and a proxy is mandatory, those other providers will instead fail after this change.

The clean contract would be:

- Scope only the proxy returned automatically by Electron's `session.resolveProxy()` to the embedded CLIProxy.
- Preserve explicitly inherited `HTTP_PROXY`, `HTTPS_PROXY`, and `ALL_PROXY` values for the whole sidecar, together with `NO_PROXY`, because those variables represent the user's explicit choice to proxy all traffic.

This keeps the intended protection against a broken OS proxy without removing the existing explicit all-traffic proxy path.

desktop/src/main/sidecar/childEnvironment.ts
jialudev: Addressed in 145baa15. The Electron-resolved proxy remains scoped to the embedded CLIProxy, while explicitly inherited HTTP_PROXY, HTTPS_PROXY, ALL_PROXY, and NO_PROXY values are now preserved for the whole sidecar so general provider traffic continues to honor the user-provided proxy configuration. Regression coverage was updated for this compatibility path.

xxx7xxxx (approved):

PR #523, opened 2026-08-26
fix(cli): disable automatic browser launch by default

xxx7xxxx:
## Summary

- Make browser launch opt-in with `--browser` while plain `csgclaw serve` continues to print the WebUI URL.
- Keep the macOS app as the default experience for normal users, while technical and professional CLI users can open the WebUI themselves.
- Prevent repeated daily development restarts from accumulating unwanted CSGClaw tabs, with tests, completion, and documentation updated.


xxx7xxxx:
@RussellLuo I will merge it if you don't have a concern

PR #526, opened 2026-08-26
refactor(template): simplify agent manifest fields

wanghaojie124:
- remove deprecated role, image metadata, and unused environment contract fields from agent.toml
- derive built-in template roles from the registry and treat local or remote templates as workers
- preserve manager/worker behavior in internal models, API responses, filtering, and image upgrade logic
- update manifest documentation and regression coverage for legacy fields and role derivation

### DIFF
--- a/internal/template/template_manifest.go
+++ b/internal/template/template_manifest.go
@@ -29,7 +29,6 @@ type templateManifest struct {
 	SchemaVersion  string               `toml:"schema_version,omitempty"`
 	Name           string               `toml:"name"`
 	Description    string               `toml:"description,omitempty"`
-	Role           string               `toml:"role"`
 	RuntimeKind    string               `toml:"runtime_kind"`
 	Version        string               `toml:"version,omitempty"`
 	Image          templateImageSection `toml:"image"`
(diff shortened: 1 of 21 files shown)


internal/template/template_manifest.go
```diff
@@ -174,13 +131,12 @@ func validateManifest(manifest templateManifest) error {
 	return nil
 }
 
-func normalizeTemplateRuntimeOptions(runtimeKind, role string, raw map[string]any) (map[string]any, error) {
+func normalizeTemplateRuntimeOptions(runtimeKind string, raw map[string]any) (map[string]any, error) {
```
xxx7xxxx, line 134: `HubPublishSpec` now emits both `execution_mode` and `memory_mode` for every Codex worker, but this normalizer still accepts exactly one key. I reproduced `POST /api/v1/hub/templates` returning HTTP 400 with `runtime_options supports only execution_mode`, so Codex worker templates cannot be published locally or remotely after the memory feature landed. Please normalize both supported keys here, align the new manifest documentation, and add an end-to-end publish regression test.

wanghaojie124: Fixed in f2b1ee0b. The manifest normalizer now accepts and validates both `execution_mode` and `memory_mode` for Codex workers. I also updated the manifest documentation and added an end-to-end regression test covering `POST /api/v1/hub/templates`, including both options in the response and generated `agent.toml`.

xxx7xxxx (approved):

PR #529, opened 2026-08-27
feat(feishu): add per-worker lark-cli binding

GatewayJ:
## Summary
- bind each Codex worker to its Feishu bot through an isolated lark-cli config and CSGClaw-backed credential source
- configure lark-cli automatically during Feishu connection when the command is available, and expose idempotent status/init controls in the worker profile
- require users to install lark-cli themselves, validate that the command starts and generated config matches the selected AppID, and surface structured configuration warnings
- enforce one worker per Feishu Bot AppID and serialize bind/disconnect cleanup for consistent state
- add managed lark-cli instructions and align the Chinese technical design document with the implementation

## Testing
- `go test ./internal/agent ./internal/participant/feishubind ./internal/runtime/codex`
- `go test ./internal/api -run 'Test(InitAgentLarkCLI.*|ReadAgentLarkCLIStatusRequiresGeneratedConfig|AgentLarkCLIStatusMarksUnstartableCommandUnavailable|DisconnectWaitsForInFlightLarkCLIBindAndClearsResult|FinalizeFeishuRegistration(ConfiguresAvailableLarkCLIForCodexWorker|KeepsConnectionWhenLarkCLIUnavailable|ReportsLarkCLIBindFailure|RejectsBotAppIDUsedByAnotherWorker))$' -count=1`
- `pnpm typecheck`
- `pnpm exec vitest run src/hooks/workspace/useAgentController.test.ts tests/components/AgentActions.test.tsx`


internal/participant/feishubind/bind.go
```diff
@@ -105,6 +117,30 @@ func BindBot(ctx context.Context, agentSvc *agent.Service, participantSvc *parti
 	return result, nil
 }
 
+func ValidateBotAppIDExclusive(participantSvc *participant.Service, agentID, appID string) error {
```
wanghaojie124, line 122: This exclusivity check only protects callers that go through `feishubind.BindBot`. The existing generic participant POST/PATCH paths and `csgclaw participant bind` write Feishu `channel_app_config` through `participant.Service` without calling this function, so two workers can still be assigned the same AppID. That violates the new invariant and can make message routing ambiguous. Please enforce the invariant in the participant service's create/update transaction (or route every mutation path through one locked helper), with regression tests for the API and CLI paths.

internal/api/lark_cli.go
```diff
@@ -0,0 +1,1075 @@
...
+	mac := hmac.New(sha256.New, []byte(secret))
+	_, _ = mac.Write([]byte(encodedPayload))
+	return base64.RawURLEncoding.EncodeToString(mac.Sum(nil))
+}
+
+func (h *Handler) internalSourceBaseURL(r *http.Request) string {
+	if h != nil && strings.TrimSpace(h.advertiseBaseURL) != "" {
+		return strings.TrimRight(strings.TrimSpace(h.advertiseBaseURL), "/")
+	}
+	if r != nil && strings.TrimSpace(r.Host) != "" {
```
wanghaojie124: Please don't derive the persisted exec-provider endpoint from the request `Host` header. A crafted init/finalize request (or a proxy forwarding an unvalidated Host) can write an attacker-controlled base URL into the worker config; the next lark-cli secret lookup then sends its bearer token to that host, and that token authorizes reading the worker's Feishu AppSecret. Use a trusted configured/listener address, or strictly validate Host against the configured public origin before persisting it.

GatewayJ: Fixed in 11329bdb. The persisted source endpoint now comes only from the configured advertise base URL (or the fixed local API default); request Host, forwarded proto, and TLS request metadata are no longer used. Added a regression test covering a crafted Host header.

internal/participant/feishubind/bind.go
GatewayJ: The CLI BIND and host bind capabilities are inconsistent; we will consider fixing this together when aligning features for the Feishu channel in the future. the mechanism involving participants

PR #530, opened 2026-08-27
feat(codex): preserve memory in agent templates

wanghaojie124:
- snapshot and restore memory_summary.md for enabled Codex templates
- expose template memory summaries in the Hub UI and align the agent memory API
- omit disabled memory and system skills while validating staging paths
- improve community publish error feedback and deployment retry handling
- update API documentation and add backend and frontend regression coverage
image
image

### DIFF
--- a/docs/tech/api.md
+++ b/docs/tech/api.md
@@ -706,8 +706,8 @@ Templates use the following layout:
 
 `AGENTS.md` and `mcp.json` are always emitted when publishing.
 Other instruction files are optional.
-Codex templates snapshot `CODEX_HOME/memories/memory_summary.md` and restore it to the same location for a newly created agent when `memory_mode` is enabled. Templates with `memory_mode = "disabled"` neither package nor restore a memory summary. They do not copy Codex's SQLite memory pipeline state or treat memory as a workspace overlay.
-For non-Codex runtimes, optional template memories are overlaid according to that Runtime's workspace convention.
+Memory is included only when each publish request explicitly sets `include_memory: true`; runtime memory enablement alone never opts a publish into exporting memory. With that publish opt-in, Codex templates snapshot `CODEX_HOME/memories/memory_summary.md` and restore it to the same location for a newly created agent when `memory_mode` is enabled. Templates with `memory_mode = "disabled"` neither package nor restore a memory summary. They do not copy Codex's SQLite memory pipeline state or treat memory as a workspace overlay.
+With the same explicit publish opt-in, supported non-Codex runtimes preserve and overlay optional template memories according to that Runtime's workspace convention.
 During agent creation, skills are installed under `skills/`, and MCP servers from `mcp.json` are applied unless the create request explicitly supplies `mcpServers`.
 
 Codex worker templates may persist their execution mode in `agent.toml`:
--- a/docs/tech/api.md
+++ b/docs/tech/api.md
@@ -758,7 +758,7 @@ Notes:
 
 - `agent_id` is required
 - `registry` uses the default publish registry when omitted
-- `include_memory` must be explicitly set to `true` to publish the Codex memory summary; it defaults to `false` because memories may contain private conversation-derived information
+- `include_memory` must be explicitly set to `true` to publish the Codex memory summary, including when republishing an existing local template via `template_id`; it defaults to `false` because memories may contain private conversation-derived information
 - Successful publish returns `201 Created`
 
 ### `GET /api/v1/hub/templates/{id}`
--- a/docs/tech/api.md
+++ b/docs/tech/api.md
@@ -469,13 +469,13 @@ Returns the Runtime's read-only primary memory document and its enabled state.
 {
   "enabled": true,
   "ready": true,
-  "name": "MEMORY.md",
-  "location": "$CODEX_HOME/memories/MEMORY.md",
+  "name": "memory_summary.md",
+  "location": "$CODEX_HOME/memories/memory_summary.md",
   "content": "# Durable memory\n"
 }
 ```
 
-`location` is a Runtime-provided logical file location for display and diagnostics; Codex currently reports `$CODEX_HOME/memories/MEMORY.md`.
+`location` is a Runtime-provided logical file location for display and diagnostics; Codex currently reports `$CODEX_HOME/memories/memory_summary.md`.
 
 #### `PUT /api/v1/agents/{id}/memory`
 
@@ -699,12 +699,13 @@ Templates use the following layout:
   instructions/AGENTS.md
   skills/<skill>/...
   mcps/mcp.json
+  memories/memory_summary.md # optional Codex memory snapshot
   memories/MEMORY.md         # optional for non-Codex runtimes
 ```
 
 `AGENTS.md` and `mcp.json` are always emitted when publishing.
 Other instruction files are optional.
-Codex templates neither publish nor restore workspace memory files; stable template context belongs in `instructions/`, while Codex-managed memory remains under its isolated `CODEX_HOME/memories/`.
+Codex templates snapshot `CODEX_HOME/memories/memory_summary.md` and restore it to the same location for a newly created agent when `memory_mode` is enabled. Templates with `memory_mode = "disabled"` neither package nor restore a memory summary. They do not copy Codex's SQLite memory pipeline state or treat memory as a workspace overlay.
 For non-Codex runtimes, optional template memories are overlaid according to that Runtime's workspace convention.
 During agent creation, skills are installed under `skills/`, and MCP servers from `mcp.json` are applied unless the create request explicitly supplies `mcpServers`.
 
--- a/docs/tech/api.md
+++ b/docs/tech/api.md
@@ -700,7 +700,8 @@ Templates use the following layout:
   skills/<skill>/...
   mcps/mcp.json
   memories/memory_summary.md # optional Codex memory snapshot
-  memories/MEMORY.md         # optional for non-Codex runtimes
+  memories/MEMORY.md         # optional root memory for non-Codex runtimes
+  memories/memory/*.md       # optional dated memory files for non-Codex runtimes
 ```
 
 `AGENTS.md` and `mcp.json` are always emitted when publishing.
--- a/docs/tech/api.md
+++ b/docs/tech/api.md
@@ -749,7 +749,8 @@ Request body:
 ```json
 {
   "agent_id": "u-alice",
-  "registry": "local"
+  "registry": "local",
+  "include_memory": false
 }
 ```
 
@@ -757,6 +758,7 @@ Notes:
 
 - `agent_id` is required
 - `registry` uses the default publish registry when omitted
+- `include_memory` must be explicitly set to `true` to publish the Codex memory summary; it defaults to `false` because memories may contain private conversation-derived information
 - Successful publish returns `201 Created`
 
 ### `GET /api/v1/hub/templates/{id}`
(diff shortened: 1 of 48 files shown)


internal/template/template_layout.go
```diff
@@ -13,15 +13,20 @@ import (
...
+const codexTemplateMemoryStagingDir = ".csgclaw-template-memory"
 
 func materializeTemplateDir(templateRoot, runtimeKind string) (WorkspaceRef, error) {
 	if !templateLayoutExists(templateRoot) && legacyTemplateWorkspaceExists(templateRoot) {
 		return materializeLegacyTemplateWorkspace(templateRoot)
 	}
 	if strings.TrimSpace(runtimeKind) == "" {
-		if _, manifest, err := loadManifestFS(os.DirFS(templateRoot), localManifestFileName, "template"); err == nil {
-			runtimeKind = manifest.RuntimeKind
+		manifestPath := filepath.Join(filepath.Base(templateRoot), localManifestFileName)
```
xxx7xxxx: `manifestPath` is passed to an `fs.FS`, whose paths must be slash-separated on every platform. On Windows, `filepath.Join` produces `dir\agent.toml`, and `os.DirFS` rejects the backslash path; `RemoteStore.FetchWorkspace` always reaches this branch with an empty runtime kind, so remote template creation fails before materialization. Please keep the `DirFS` rooted at `templateRoot` and pass `localManifestFileName`, or construct an fs-valid path with `path.Join`/`filepath.ToSlash`, and add a Windows-oriented regression test.

web/app/src/pages/HubPage/components/HubDetailPane/HubDetailPane.tsx
```diff
@@ -853,7 +865,9 @@ export function HubDetailPane({
           ? firstTemplateSectionFile(workspaceEntries, "skills", "SKILL.md")
           : tab === "mcp"
             ? firstTemplateSectionFile(workspaceEntries, "mcps", "mcp.json")
-            : firstTemplateSectionFile(workspaceEntries, "instructions", "AGENTS.md");
+            : tab === "memory"
+              ? firstTemplateSectionFile(workspaceEntries, "memories", "memory_summary.md")
```
xxx7xxxx, line 881: In the real Hub flow, `workspaceEntries` initially contains only the root directories. Clicking Memory therefore selects an empty path before `memories/` is lazily loaded; after that listing arrives, nothing reselects or preloads `memory_summary.md` (the controller preload list covers instructions, MCPs, and skills only), so the panel shows the empty state even when the file exists. Please select/preload the memory file after the directory listing becomes available and make the test model the lazy loading sequence.

internal/agent/service.go
```diff
@@ -247,6 +247,17 @@ func (s *Service) HubPublishSpec(agentID string) (hub.PublishSpec, error) {
 	if err != nil {
 		return hub.PublishSpec{}, err
 	}
+	memoryPath := ""
+	if runtimeOptions[templateMemoryModeKey] != templateMemoryModeDisabled {
```
xxx7xxxx: This makes every memory-enabled Codex publish include the learned memory summary, and memory is enabled by default. The local and community publish dialogs still describe only template metadata and provide no disclosure or per-publish choice, so publishing an existing agent can upload conversation-derived private facts to the Hub without informed consent. Please make memory export explicit, such as a visible opt-in/confirmation with a safe default, and cover the remote-publish flow.

web/app/tests/shared/i18n.test.ts
```diff
@@ -51,10 +51,21 @@ describe("i18n messages", () => {
...
         createTranslator("en"),
       ),
-    ).toContain("same name already exists");
+    ).toContain("community repository already exists");
+  });
+
+  it("preserves the upstream reason for generic template publishing failures", () => {
+    const message =
+      'publish hub template to "official": remote hub request failed with status 500: failed to update repository path';
+    expect(
```
xxx7xxxx: This new assertion is not in the repository's Prettier form, so `./scripts/web-pnpm.sh check` fails on this PR-touched file. Please format the test before merging.

internal/template/template_layout.go
wanghaojie124: Fixed in e7da6bcd. The manifest path is now constructed as an `fs.FS`-valid slash-separated path, while preserving the existing parent-rooted `DirFS` safety boundary. I also added a Windows-only regression test and verified the template package cross-compiles for Windows amd64.

web/app/src/pages/HubPage/components/HubDetailPane/HubDetailPane.tsx
wanghaojie124: Fixed in e7da6bcd. When the Memory tab is active, the pane now reselects `memories/memory_summary.md` as soon as the lazily loaded directory entry becomes available. The regression test now starts with root directories only and models the lazy-loading sequence.

internal/agent/service.go
wanghaojie124: Fixed in e7da6bcd. Memory export now requires an explicit, default-off checkbox in both local and community publish dialogs, with a privacy warning. The API also defaults `include_memory` to false, so older or direct clients cannot upload memory implicitly. Added frontend coverage and an API-level publish test for both opt-out and opt-in behavior.

web/app/tests/shared/i18n.test.ts
wanghaojie124: Fixed in e7da6bcd. The touched test is now in Prettier form. `./scripts/web-pnpm.sh check` passes with all 956 frontend tests.

internal/api/handler.go
```diff
@@ -1756,7 +1756,7 @@ func (h *Handler) handleHubTemplates(w http.ResponseWriter, r *http.Request) {
 			item, err = hubSvc.PublishTemplate(r.Context(), req.TemplateID, req.Registry)
 		} else {
 			var spec hub.PublishSpec
-			spec, err = h.svc.HubPublishSpec(req.AgentID)
+			spec, err = h.svc.HubPublishSpec(req.AgentID, req.IncludeMemory)
```
xxx7xxxx, line 1759: The opt-in is still bypassed when publishing an existing local template. The `template_id` branch above calls `PublishTemplate` without `IncludeMemory`, and `PublishTemplate` carries the fetched workspace's `MemoryPath` into the official registry. A memory-bearing local template can therefore upload conversation-derived memory to the community with no memory choice or warning in the Hub publish dialog. Please make template-to-community memory transfer explicitly opt-in and default-off too, and cover that route with an API/UI regression test.

web/app/src/pages/AgentPage/components/AgentDetailPane/AgentDetailPane.tsx
```diff
@@ -1229,6 +1237,19 @@ export const AgentDetailPane = forwardRef<AgentDetailPaneHandle, AgentDetailPane
                 onChange={(event) => setPublishTemplateDescription(event.target.value)}
               />
             </label>
+            {runtimeKind === "codex" ? (
```
xxx7xxxx: The backend now gates non-Codex workspace memory behind `IncludeMemory` too, but this checkbox is rendered only for Codex. OpenClaw agents are still allowed to save local templates, so every UI publish now sends `false` and silently drops their existing `memory/` or `MEMORY.md` content with no way to opt in. Please either keep the new gate Codex-specific or expose the same explicit memory choice for every runtime whose template memory is supported, and add an OpenClaw UI regression test.

internal/api/handler.go
wanghaojie124: Fixed in 6d3be9ae. Existing local-template publishes now forward the explicit, default-off include_memory choice through PublishTemplate, and both LocalStore and RemoteStore enforce that policy again at the storage boundary. The packager also always excludes the reserved Codex memory staging directory, so memory cannot leak through instructions. Added service/API/UI and remote-archive regression coverage for opt-out and opt-in behavior.

web/app/src/pages/AgentPage/components/AgentDetailPane/AgentDetailPane.tsx
wanghaojie124: Fixed in 97e93c95. The publish dialog now exposes the same default-off memory checkbox for OpenClaw local-template publishing and forwards true only after explicit opt-in. OpenClaw MEMORY.md and memory/* are also preserved and restored to their original workspace paths, with compatibility for previously flattened templates. Added OpenClaw UI, template round-trip, legacy-layout, and agent-creation regression tests; the frontend check passes with all 957 tests.

internal/apitypes/hub.go
```diff
@@ -3,12 +3,13 @@ package apitypes
...
-	Registry    string  `json:"registry,omitempty"`
-	Name        string  `json:"name,omitempty"`
-	Description *string `json:"description,omitempty"`
-	Deploy      bool    `json:"deploy,omitempty"`
+	AgentID       string  `json:"agent_id,omitempty"`
+	TemplateID    string  `json:"template_id,omitempty"`
+	Registry      string  `json:"registry,omitempty"`
+	Name          string  `json:"name,omitempty"`
+	Description   *string `json:"description,omitempty"`
+	IncludeMemory bool    `json:"include_memory,omitempty"`
```
xxx7xxxx, line 11: `include_memory` is exposed through the API and web flows, but `csgclaw template publish` has no corresponding flag and always leaves this false. CLI users therefore cannot create a memory-bearing Codex or OpenClaw template. Please add an explicit `--include-memory` flag with an appropriate sensitive-data warning and CLI coverage, or document that this feature is intentionally unavailable from the CLI.

docs/tech/api.md
```diff
@@ -699,12 +699,14 @@ Templates use the following layout:
...
-  memories/MEMORY.md         # optional for non-Codex runtimes
+  memories/memory_summary.md # optional Codex memory snapshot
+  memories/MEMORY.md         # optional root memory for non-Codex runtimes
+  memories/memory/*.md       # optional dated memory files for non-Codex runtimes
 ```
 
 `AGENTS.md` and `mcp.json` are always emitted when publishing.
 Other instruction files are optional.
-Codex templates neither publish nor restore workspace memory files; stable template context belongs in `instructions/`, while Codex-managed memory remains under its isolated `CODEX_HOME/memories/`.
+Codex templates snapshot `CODEX_HOME/memories/memory_summary.md` and restore it to the same location for a newly created agent when `memory_mode` is enabled. Templates with `memory_mode = "disabled"` neither package nor restore a memory summary. They do not copy Codex's SQLite memory pipeline state or treat memory as a workspace overlay.
```
xxx7xxxx: This overview still says that enabling `memory_mode` causes Codex memory to be snapshotted, and the next sentence describes non-Codex memory without mentioning the publish gate. After this PR, neither path includes memory unless `include_memory: true` is supplied, so this conflicts with the later API note and can mislead clients into thinking runtime memory enablement is sufficient. Please state the explicit publish opt-in here for both Codex and supported non-Codex runtimes, and mirror it in the Chinese document.

internal/apitypes/hub.go
wanghaojie124: Fixed in f44e394d. `csgclaw template publish` now exposes an explicit, default-off `--include-memory` flag. When selected it sends `include_memory: true` and prints a sensitive-data warning to stderr; without the flag the field remains omitted. CLI and shell-completion regression tests cover both the opt-in and safe default.

docs/tech/api.md
wanghaojie124: Fixed in f44e394d. The English and Chinese API overviews now state that runtime memory enablement never authorizes export by itself: every Codex or supported non-Codex publish must explicitly set `include_memory: true`. The existing `memory_mode = disabled` behavior remains documented separately.

xxx7xxxx (approved):

PR #531, opened 2026-08-27
refactor(agent): route administration through agent engine

xxx7xxxx:
## Summary

- Route Agent administration through a lean Agent Engine resource interface.
- Preserve lifecycle, subresource, secret, persistence, and restart behavior with end-to-end regression coverage.

### DIFF
--- a/internal/agentengine/agent_facade.go
+++ b/internal/agentengine/agent_facade.go
@@ -166,7 +166,11 @@ func (f agentFacade) Get(ctx context.Context, agentID string, options AgentGetOp
 		if !ok {
 			return Agent{}, &TurnError{Code: ErrorAgentNotFound, Message: fmt.Sprintf("agent %q not found", agentID)}
 		}
-		selected = f.withAdoptedMCPServers(ctx, selected)
+		var err error
+		selected, err = f.withAdoptedMCPServers(ctx, selected, options.AdoptMCPServers)
+		if err != nil {
+			return Agent{}, err
+		}
 		if options.IncludeDocuments {
 			return f.convertWithDocuments(ctx, selected)
 		}
@@ -179,22 +183,27 @@ func (f agentFacade) Get(ctx context.Context, agentID string, options AgentGetOp
 	if !ok {
 		return Agent{}, &TurnError{Code: ErrorAgentNotFound, Message: fmt.Sprintf("agent %q not found", agentID)}
 	}
-	selected = f.withAdoptedMCPServers(ctx, selected)
+	var err error
+	selected, err = f.withAdoptedMCPServers(ctx, selected, options.AdoptMCPServers)
+	if err != nil {
+		return Agent{}, err
+	}
 	if options.IncludeDocuments {
 		return f.convertWithDocuments(ctx, selected)
 	}
 	return f.convert(selected)
 }
 
-func (f agentFacade) withAdoptedMCPServers(ctx context.Context, selected agent.Agent) agent.Agent {
-	if selected.MCPServers != nil {
-		return selected
+func (f agentFacade) withAdoptedMCPServers(ctx context.Context, selected agent.Agent, adopt bool) (agent.Agent, error) {
+	if !adopt || selected.MCPServers != nil {
+		return selected, nil
 	}
 	view, err := f.service.MCPServersView(ctx, selected.ID)
-	if err == nil {
-		selected.MCPServers = view.Servers
+	if err != nil {
+		return agent.Agent{}, err
 	}
-	return selected
+	selected.MCPServers = view.Servers
+	return selected, nil
 }
 
 func (f agentFacade) List(ctx context.Context, options AgentListOptions) ([]Agent, error) {
--- a/internal/agentengine/agent_facade.go
+++ b/internal/agentengine/agent_facade.go
@@ -2,6 +2,7 @@ package agentengine
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"maps"
 	"reflect"
@@ -118,6 +119,17 @@ func (f agentFacade) Create(ctx context.Context, request AgentCreateRequest) (Ag
 			return Agent{}, err
 		}
 	}
+	if spec.Memory != nil {
+		if _, err := f.service.UpdateMemoryEnabled(ctx, created.ID, spec.Memory.Enabled); err != nil {
+			if createdByCall {
+				if rollbackErr := f.service.Delete(context.Background(), created.ID); rollbackErr != nil {
+					return Agent{}, errors.Join(err, fmt.Errorf("rollback Agent after memory reconciliation failure: %w", rollbackErr))
+				}
+			}
+			return Agent{}, err
+		}
+		created, _ = f.service.Agent(created.ID)
+	}
 	if spec.DesiredState == AgentDesiredStateStopped {
 		created, err = f.service.Stop(ctx, created.ID)
 		if err != nil {
@@ -128,7 +140,15 @@ func (f agentFacade) Create(ctx context.Context, request AgentCreateRequest) (Ag
 	if err != nil {
 		return Agent{}, err
 	}
-	return f.convert(created)
+	converted, err := f.convert(created)
+	if err != nil {
+		return Agent{}, err
+	}
+	if spec.Memory != nil {
+		memory := *spec.Memory
+		converted.Spec.Memory = &memory
+	}
+	return converted, nil
 }
 
 func (f agentFacade) Get(ctx context.Context, agentID string, options AgentGetOptions) (Agent, error) {
@@ -194,7 +214,7 @@ func (f agentFacade) Update(ctx context.Context, agentID string, request AgentUp
 	loadSkills := len(request.FieldMask) == 0 || fieldMaskContains(request.FieldMask, "skills")
 	loadMemory := fieldMaskContains(request.FieldMask, "memory") || request.Spec.Memory != nil
 	reconcileDesiredState := fieldMaskContains(request.FieldMask, "desired_state")
-	return f.updateDesired(ctx, agentID, request.ResourceVersion, loadSkills, loadMemory, reconcileDesiredState, func(current AgentSpec) (AgentSpec, error) {
+	return f.updateDesired(ctx, agentID, request.ResourceVersion, loadSkills, loadMemory, reconcileDesiredState, false, func(current AgentSpec) (AgentSpec, error) {
 		return mergeAgentUpdate(current, request), nil
 	})
 }
@@ -239,6 +259,8 @@ func mergeAgentUpdate(current AgentSpec, request AgentUpdateRequest) AgentSpec {
 			next.Runtime.InitShell = request.Spec.Runtime.InitShell
 		case "model":
 			next.Model = request.Spec.Model
+		case "model.selector":
+			next.Model.Selector = request.Spec.Model.Selector
 		case "skills":
 			next.Skills = append([]string(nil), request.Spec.Skills...)
 		case "mcp_servers":
@@ -260,7 +282,8 @@ func mergeAgentUpdate(current AgentSpec, request AgentUpdateRequest) AgentSpec {
 type agentSpecChange struct {
 	name, description, instructions, image bool
 	runtimeSelection, runtimeOptions       bool
-	model, skills, mcpServers              bool
+	modelSelector, modelConfig             bool
+	skills, mcpServers                     bool
 	memory                                 bool
 	runtimeCredentials, runtimeInitShell   bool
 	desiredState                           bool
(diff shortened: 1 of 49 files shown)


internal/api/handler.go
```diff
@@ -1584,13 +1609,28 @@ func (h *Handler) handleCreateAgentWorker(w http.ResponseWriter, r *http.Request
...
+	ctx := agentengine.WithLocalTemplateService(r.Context(), hubSvc)
+	agents := h.agentEngine.Agents()
+	var created agentengine.Agent
+	if createReq.Replace {
+		current, getErr := agents.Get(ctx, createReq.Spec.ID, agentengine.AgentGetOptions{})
+		if getErr != nil {
+			writeAgentOperationErrorWithBillingURL(w, getErr, http.StatusBadRequest, llm.OpenCSGBillingURL(createReq.Spec.AgentProfile))
+			return
+		}
+		created, err = agents.Update(ctx, current.ID, engineReplaceRequest(createReq, current))
```
wanghaojie124: `replace=true` is now routed through `Update`, so it no longer preserves the existing recreate contract. In particular, the CLI intentionally sends `field_mask:["id"]` for an in-place recreate; `engineReplaceRequest` drops `id`, producing an empty field mask, which `AgentInterface.Update` interprets as a complete replacement from an otherwise mostly empty spec. This can clear existing configuration, and even replace requests with other masked fields only patch the resource instead of rebuilding its runtime. Please keep the replace path mapped to a recreate-capable engine operation (while applying the merged fields), and add an API regression test for the id-only CLI payload.

internal/api/handler.go
```diff
@@ -1136,29 +1142,35 @@ func (h *Handler) handleAgentByID(w http.ResponseWriter, r *http.Request) {
...
-		}
 		// A detail request is an explicit inspection point. Unlike the roster
 		// path, it may make the bounded optional readiness probe needed to tell
 		// whether a running gateway can accept work.
 		ctx, cancel := context.WithTimeout(r.Context(), agentRuntimeStatusTimeout)
 		defer cancel()
-		a, ok := h.svc.Inspect(ctx, id)
-		if !ok {
+		item, err := agents.Get(ctx, id, agentengine.AgentGetOptions{Reload: true, ProbeRuntime: true})
+		if err != nil {
```
wanghaojie124, line 1151: This maps every Engine error to 404, including failures from `Reload` and the bounded runtime probe. The previous handler returned 500 when reload failed, so storage/engine outages will now be reported as a missing agent and may trigger incorrect client behavior. Please distinguish not-found from operational errors (ideally via typed Engine error codes) and preserve 5xx for reload/probe failures.

internal/api/agent_engine_admin.go
```diff
@@ -0,0 +1,298 @@
...
+		result.Spec.Runtime.InitShell = *request.RuntimeInitShell
+		add("runtime.init_shell")
+	}
+	if request.AgentProfile != nil {
+		result.Spec.Model = engineModelFromProfile(*request.AgentProfile)
+		add("model")
+	}
+	if request.Profile != nil {
+		result.Spec.Model.Selector = *request.Profile
+		add("model")
```
wanghaojie124: A selector-only PATCH is being promoted to a full `model` replacement. For a request such as `{"profile":"codex-fast"}`, the translated `ModelSpec` contains only `Selector`; `mergeAgentUpdate` then replaces the entire current model, and `preserveWriteOnlyFields` restores only the API key. This clears provider/base URL/model ID/headers/env/options, whereas the previous service path updated only `Agent.Profile`. Please add a selector-specific field mask/merge path, or merge the selector into the current model, with an end-to-end regression test that verifies the rest of the model configuration is preserved.

internal/participant/service.go
```diff
@@ -540,7 +562,7 @@ func (s *Service) RepairDanglingCSGClawAgentParticipants() ([]apitypes.Participa
 		}
 		agentID := strings.TrimSpace(item.AgentID)
 		if agentID != "" {
-			if _, ok := s.agents.Agent(agentID); ok {
+			if _, err := s.agents.Get(context.Background(), agentID, agentengine.AgentGetOptions{}); err == nil {
```
wanghaojie124, line 565: Any Engine `Get` error is treated as proof that the agent is absent, after which this repair path deletes the participant and potentially its local IM user. With the new transport-capable Engine boundary, a timeout or temporary backend/storage failure is therefore destructive. Please delete only on a typed not-found result and return/propagate operational errors without mutating participant state; a contract test with a failing AgentInterface would protect this path.

internal/agentengine/agent_facade.go
```diff
@@ -43,55 +54,131 @@ type agentFacade struct {
...
+				managerSpec.MCPServers = spec.MCPServers
+			}
+			managerSpec.DesiredState = spec.DesiredState
+			created, err = f.service.Update(ctx, created.ID, updateAgentRequestForRole(managerSpec, spec.Role))
+			if err != nil {
+				return Agent{}, err
+			}
+		}
+	}
+	if spec.Skills != nil {
```
wanghaojie124, line 114: `AgentSpec.Memory` is part of the desired resource contract, but Create never applies it. A caller can request memory disabled/enabled and receive a successfully created agent whose memory state does not match the submitted spec. Please reconcile `spec.Memory` during Create (with rollback if the subresource update fails), and cover it in the shared AgentInterface contract tests.

internal/api/handler.go
xxx7xxxx: Fixed in 8199aefe. Replace now passes the field-masked desired-state update through Recreate under the Agent lifecycle gate. The id field remains in the mask, so the CLI id-only payload preserves the complete spec while still rebuilding the Runtime. Added API and live CLI regressions.

internal/api/handler.go
xxx7xxxx: Fixed in 8199aefe. Detail GET now maps only typed agent_unavailable errors to 404; reload and other operational Engine failures remain 500. Added explicit not-found and storage-failure coverage.

internal/api/agent_engine_admin.go
xxx7xxxx: Fixed in 8199aefe. Selector changes now use a dedicated model.selector field mask and merge only that field. Provider, base URL, model ID, headers, environment, request options, and credentials remain unchanged, with shared contract, API, and live E2E coverage.

internal/participant/service.go
xxx7xxxx: Fixed in 8199aefe. Participant repair now deletes only when AgentInterface returns typed agent_unavailable. Operational Engine errors are propagated before any participant or IM-user mutation. Added a failing-AgentInterface regression test.

internal/agentengine/agent_facade.go
xxx7xxxx: Fixed in 8199aefe. Create now reconciles AgentSpec.Memory, returns the applied memory intent, and deletes the newly created Agent and Runtime if reconciliation fails. The behavior is covered by the shared AgentInterface contract and an explicit rollback regression.

internal/api/handler.go
wanghaojie124: Thanks for the follow-up. This is improved for untyped reload/storage errors, but the original issue is not fully resolved because `ErrorAgentUnavailable` still represents both not-found and service/backend unavailability. For example, `unavailableAgents.Get` and `agentFacade.Get` with a nil service return `ErrorAgentUnavailable` with "agent service is unavailable", while missing agents return the same code. `writeAgentGetError` therefore still maps a genuine Engine outage to 404. Please introduce a distinct `ErrorAgentNotFound` (or equivalent typed distinction), map only that code to 404, and add a regression where `Get` returns `ErrorAgentUnavailable` and the endpoint remains 5xx.

internal/participant/service.go
wanghaojie124: Thanks for adding the operational-error guard. This remains unsafe for failures reported as `ErrorAgentUnavailable`: that code is currently returned both when an agent is missing and when the Agent service/backend itself is unavailable (`unavailableAgents.Get` is a concrete example). The repair path still treats either case as permission to delete the participant and local IM user. Please delete only on a dedicated `ErrorAgentNotFound` code, propagate `ErrorAgentUnavailable`, and add a regression test whose fake `Get` returns `ErrorAgentUnavailable` and verifies that neither participant nor IM-user state is mutated.

internal/api/handler.go
xxx7xxxx: Fixed in 6ea83144. Agent resource lookup now has a dedicated agent_not_found code. Real Engine and MemoryClient return it only for missing resources; agent_unavailable remains reserved for service/backend outages. Detail GET maps only agent_not_found to 404, with explicit regressions proving agent_unavailable and untyped storage failures remain 500.

internal/participant/service.go
xxx7xxxx: Fixed in 6ea83144. Participant repair now deletes only on the dedicated agent_not_found code. Both untyped operational errors and typed agent_unavailable errors are propagated before participant or IM-user mutation, with regression coverage for each case.

internal/api/model_providers.go
```diff
@@ -337,14 +338,15 @@ func (h *Handler) modelProviderInUse(llm config.LLMConfig, id string) bool {
...
-		if agent.NormalizeModelProviderID(item.AgentProfile.ModelProviderID) == id {
-			return true
-		}
-		if agent.SelectorUsesModelProvider(item.Profile, id) {
+	items, err := h.agentEngine.Agents().List(context.Background(), agentengine.AgentListOptions{})
+	if err != nil {
+		return false
+	}
+	for _, item := range items {
+		if agent.NormalizeModelProviderID(item.Spec.Model.ProviderID) == id {
```
wanghaojie124, line 349: The Engine migration dropped the selector-based usage check that existed on `item.Profile`. `ModelSpec.Selector` can intentionally differ from `ProviderID` (the new `model.selector` PATCH preserves the existing model config), so an agent may select `provider-b.model-b` while this loop only sees the previous/configured ProviderID. Deleting provider-b will then incorrectly pass the in-use guard and remove a provider still referenced by the agent. Please also check `agent.SelectorUsesModelProvider(item.Spec.Model.Selector, id)` and add a regression covering a selector-only agent update followed by provider deletion.

internal/api/agent_mcp_servers.go
```diff
@@ -76,17 +80,48 @@ func (h *Handler) handleBatchAddAgentMCPServers(w http.ResponseWriter, r *http.R
...
 		return
 	}
-	updated, err := h.svc.AddMCPServersFromHub(r.Context(), id, req.Names, servers)
-	if err != nil {
-		writeAgentMCPServersMutationError(w, err)
-		return
+	agents := h.agentEngine.Agents()
+	current, err := agents.Get(r.Context(), id, agentengine.AgentGetOptions{})
+	if err == nil {
+		if current.Spec.MCPServers == nil {
```
wanghaojie124, line 86: When persisted `MCPServers` is nil, this now starts from an empty map. The previous management path deliberately called `currentMCPServersForManagement` to import every server from the Runtime on the first mutation; therefore adding one catalog server here can replace the Runtime configuration with only that new server and silently drop all pre-existing unmanaged entries (and GET no longer exposes them either). Please preserve the initial Runtime-adoption behavior behind the Agent resource/Engine boundary, and add a regression where an unmanaged Runtime already has one server before batch-add/delete.

internal/api/model_providers.go
xxx7xxxx: Fixed in 7d90409d. Provider usage now checks both Model.ProviderID and Model.Selector. The regression performs a selector-only Agent PATCH to reference the provider while retaining a different configured ProviderID, then verifies provider deletion is rejected.

internal/api/agent_mcp_servers.go
xxx7xxxx: Fixed in 7d90409d. Agent Engine Get now adopts Runtime MCP servers when persisted MCPServers is nil. GET exposes the adopted map, and the first batch add/delete starts from that complete state before persistence, so unmanaged entries are preserved. The regression verifies adoption, add, persistence, delete, and no subsequent re-import.

internal/agentengine/agent_facade.go
```diff
@@ -43,55 +55,163 @@ type agentFacade struct {
...
 	}
 	return f.convert(selected)
 }
 
-func (f agentFacade) List(ctx context.Context) ([]Agent, error) {
+func (f agentFacade) withAdoptedMCPServers(ctx context.Context, selected agent.Agent) agent.Agent {
+	if selected.MCPServers != nil {
+		return selected
+	}
+	view, err := f.service.MCPServersView(ctx, selected.ID)
```
wanghaojie124, line 201: This makes every ordinary Agent `Get` perform Runtime MCP I/O whenever persisted MCP state is nil, and then silently discards any read failure. Besides leaving batch mutations able to overwrite an unreadable Runtime from an empty map, unrelated callers such as team directories, task identity lookup, session resolution, and participant repair now invoke this path—many with `context.Background()` and no timeout—so a stuck Runtime MCP controller can indefinitely block otherwise metadata-only operations. Please make adoption an explicit/error-returning operation used by the MCP management path (or an explicit Get option), while keeping normal resource reads independent of Runtime MCP availability; mutation must abort if initial adoption fails.

xxx7xxxx: Fixed in 32183cca. Runtime MCP adoption is now opt-in through AgentGetOptions and used only by the MCP management routes, so ordinary Agent reads no longer call Runtime MCP controllers. Initial adoption failures are returned to the caller, and batch add/delete abort without overwriting unmanaged state. Added regressions for metadata-read independence, successful adoption, and failed mutation preservation; validated end to end.

wanghaojie124:
LGTM

PR #533, opened 2026-08-28
fix(web): prevent agent profile drawer flicker after delete

youngbeom-shin:
## Summary

- close the conversation profile drawer as soon as its agent is successfully deleted
- clear the stored drawer trigger so stale agent-list responses cannot reopen the deleted profile
- keep the drawer open when deletion fails so the error remains visible
- add regression coverage for successful and failed deletion callbacks


jialudev (approved):
LGTM

PR #537, opened 2026-08-31
feat(knowledgebase): integrate AgenticHub LLM Wiki

jialudev:
- Integrate AgenticHub LLM Wiki resources into CSGClaw so signed-in users can discover eligible knowledge bases, add their MCP servers, and bind them to agents.
- Preserve knowledge-base MCP bindings in templates while resolving the current AgenticHub endpoint and runner credentials when agents are restored.
- Use AgenticHub-provided MCP endpoints and remove development mock resources from the knowledge-base flow.


internal/agent/knowledge_mcp.go
```diff
@@ -0,0 +1,10 @@
+package agent
+
+import "csgclaw/internal/knowledgebase"
+
+func (s *Service) materializeRuntimeMCPServers(runtimeKind string, servers map[string]any) (map[string]any, error) {
+	if servers == nil {
+		return cloneMCPServers(servers), nil
+	}
+	return knowledgebase.RuntimeServers(servers)
```
xxx7xxxx, line 13: This runtime materialization path only removes the management metadata; it keeps the URL and Authorization header captured when the knowledge base was first added. After the OpenCSG login/environment or AgenticHub endpoint changes, validation can use the current connection while a restarted agent still receives stale values and cannot connect. Please hydrate each managed server with the current connection here, or persist reference-only metadata and resolve it during materialization.

web/app/src/pages/WorkspacePage/components/WorkspaceSidebar/KnowledgeBaseDiscoveryDialog.tsx
```diff
@@ -0,0 +1,156 @@
...
+        <DialogBody className={styles.body}>
+          <label className={styles.search}>
+            <Search size={18} strokeWidth={1.8} aria-hidden="true" />
+            <TextInput
+              className={styles.searchInput}
+              type="search"
+              value={search}
+              placeholder={t("resourcesKnowledgeBaseSearchPlaceholder")}
+              aria-label={t("resourcesKnowledgeBaseSearchPlaceholder")}
+              onChange={(event) => setSearch(event.currentTarget.value)}
```
xxx7xxxx: This filters only the already-loaded page, while both knowledge-base queries request page 1 with `per=50`. The selection hook already has a debounced remote-search state, but this dialog never updates it, so a knowledge base outside the first 50 cannot be found or added, and configured entries outside that page disappear from the sidebar. Please wire this input to the remote query and support paging through the result set.

web/app/src/hooks/workspace/useWorkspaceController.ts
```diff
@@ -717,6 +724,32 @@ export function useWorkspaceController() {
...
+    try {
+      document = JSON.parse(hub.mcpCreateInitialDocument);
+    } catch {
+      return;
+    }
+    const payload = mcpServerPayloadFromDocument(document);
+    if (!payload?.name) {
+      return;
+    }
+    navigatePane({ type: WorkspacePaneTypes.hub, id: payload.name, resourceType: "mcp" }, rooms);
```
xxx7xxxx: This navigates to the prospective MCP name before the server exists. If the user cancels the dialog or saving fails, the URL remains on that missing resource while `selectedMCPServer` falls back to the first installed server, so the route and displayed detail disagree. Please navigate only after creation succeeds, using the saved name.

jialudev: Fixed in 78fc1b65. The eager navigation on dialog open was removed. The create handler now navigates to payload.name only after onCreateMCP returns success; cancel and save failure leave the current route unchanged.

internal/agent/knowledge_mcp.go
jialudev: Fixed in 78fc1b65. Runtime materialization now hydrates each managed knowledge-base server from the current connection before removing management metadata. The regression test starts from the originally published stale URL and token and verifies that both are refreshed.

web/app/src/pages/WorkspacePage/components/WorkspaceSidebar/KnowledgeBaseDiscoveryDialog.tsx
jialudev: Fixed in 78fc1b65. The dialog search now drives the debounced remote query, discovery results support incremental paging, and the unfiltered catalog automatically loads all pages so configured entries beyond the first 50 remain visible in the sidebar.

internal/agent/knowledge_mcp.go
```diff
@@ -0,0 +1,18 @@
...
+	"context"
+
+	"csgclaw/internal/knowledgebase"
+)
+
+func (s *Service) materializeRuntimeMCPServers(ctx context.Context, runtimeKind string, servers map[string]any) (map[string]any, error) {
+	if servers == nil {
+		return cloneMCPServers(servers), nil
+	}
+	hydrated, err := knowledgebase.HydrateTemplateServers(ctx, servers)
```
xxx7xxxx: Hydrating every materialization also hydrates the previous snapshot used by `reconcileMCPServers`. If the user has signed out, lost access, or the knowledge base is unavailable, deleting that managed server now returns the hydration error before the runtime reconciliation runs. At that point `update` has already persisted the desired deletion, so the API reports failure while persisted state and the live runtime diverge. Please hydrate only the current desired servers and materialize the previous snapshot from its persisted runtime values so removals remain possible without current upstream access.

jialudev: Addressed in fa01cb57. Runtime materialization no longer contacts AgenticHub. Both the current desired servers and the previous snapshot used by runtime reconciliation are materialized from their persisted URL and Authorization values, with only CSGClaw management metadata removed. This keeps deletion possible when the user is signed out, has lost access, or the knowledge base is temporarily unavailable. Source refresh is now an explicit user action through the MCP or Agent UI. TestDeleteManagedKnowledgeBaseMCPUsesPersistedSnapshotWithoutSourceAccess covers the removal path without upstream access.

jialudev: Some additional context on the final design after considering this thread together with the first review comment (`discussion_r3900019240`):

The first fix resolved every managed knowledge-base MCP from AgenticHub during runtime materialization. This thread correctly exposed the lifecycle problem with that approach: deleting, reconciling, restarting, or otherwise managing an existing Agent would then depend on the current login and upstream availability.

The final design treats the Agent's MCP configuration as a self-contained runnable snapshot. The Agent is also the unit that users save and publish, so keeping its MCP configuration independent from the local global MCP catalog avoids introducing persistent catalog/AgenticHub reference dependencies into Agent, template, and Runtime lifecycles. Existing Agents therefore keep using their saved URL and token; source changes are detected and applied only through the explicit check/update action. Updating from the Agent page refreshes both the global MCP copy and that Agent's copy.

Community publication is handled at a separate boundary: credentials are removed from the published template while the knowledge-base identity metadata is retained. When another user creates an Agent from the template, CSGClaw resolves that identity once and creates a new runnable snapshot with that user's current endpoint and token. This knowledge-base bridge is also expected to move toward the plugin/connector model later, so the current implementation deliberately avoids making a temporary reference mechanism a permanent cross-layer contract.

I documented the complete list -> global MCP -> Agent -> community template -> new Agent flow, including the manual update behavior, here: [AgenticHub knowledge-base MCP flow](https://github.com/jialudev/csgclaw/blob/297dfbfc4572c3a01d46264fa74f6f88f5ad11fb/docs/tech/knowledge-base-mcp.zh.md).

internal/api/mcp_server_source.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		writeMCPServerSourceError(w, err)
+		return
+	}
+	resolved.Status.GlobalServerName = name
+	switch r.Method {
+	case http.MethodGet:
+		writeJSON(w, http.StatusOK, resolved.Status)
+	case http.MethodPost:
+		state, err := h.mcp.UpdateServer(r.Context(), name, name, resolved.CanonicalConfig)
```
xxx7xxxx: This sync persists `CanonicalConfig`, which is rebuilt from the AgenticHub item by `ServerConfig`. As a result, local MCP runtime settings that are not source-owned are removed. I reproduced this by syncing a managed server with `startup_timeout_sec` and `tool_timeout_sec`; both fields disappeared from the saved config. `RefreshManagedServerSnapshot` already returns `Refreshed`, which preserves the existing local fields while replacing the endpoint and credentials. Please persist that refreshed snapshot here and in the Agent sync path so a source update does not erase local runtime tuning.

GatewayJ (approved):

internal/api/mcp_server_source.go
jialudev: Fixed in ab4d36cf. Global and Agent snapshots are now refreshed independently: each keeps its own local runtime fields while only the AgenticHub-owned endpoint, transport, and Authorization are updated. Existing global entries are merged from their own persisted snapshot; CanonicalConfig is used only when a global entry does not exist yet. The header merge also replaces Authorization case-insensitively without removing unrelated custom headers. Targeted tests cover startup_timeout_sec, tool_timeout_sec, and distinct global/Agent tuning values. The snapshot/update flow is documented in docs/tech/knowledge-base-mcp.zh.md.

xxx7xxxx (approved):

PR #538, opened 2026-08-31
feat(chat): add generated file delivery and previews

xxx7xxxx:
## Summary

- deliver agent-generated files as durable chat attachments
- add secure multi-format previews with syntax highlighting and responsive controls

## Screenshots

image

image

image


internal/runtime/codex/appserver_manager.go
```diff
@@ -412,6 +412,9 @@ func (m *appServerManager) ensureSession(ctx context.Context, handle SessionHand
 
 	live.mu.Lock()
 	if threadID := strings.TrimSpace(live.conversationSessions[conversationKey]); threadID != "" && live.loadedConversations[conversationKey] {
+		if publishFiles {
```
wanghaojie124: This upgrades every already-loaded mapping to `filePublishingThreads=true` without knowing how that thread was created. A thread created through legacy `EnsureSession` did not receive `csgclaw_publish_file` in `dynamicTools`, so later calling `EnsureEngineSession` marks it file-capable even though the app-server thread cannot advertise the tool. The same ambiguity exists for mappings persisted before this feature. Please persist/verify the capability provenance, or start a file-capable thread when the existing mapping is not known to retain the tool; add a regression test that calls legacy `EnsureSession` before `EnsureEngineSession`.

web/app/src/pages/ConversationPage/components/ConversationPane/ConversationPane.tsx
```diff
@@ -350,7 +366,15 @@ export function ConversationPane({
...
-  const sidePanel = agentDetailPanel ?? activityPanel ?? threadPanel;
+  const documentPreviewPanel = documentPreview ? (
+    <DocumentPreviewPanel
+      {...documentPreview}
+      t={t}
+      onClose={() => setDocumentPreview(null)}
+      onIndexChange={(index) => setDocumentPreview((current) => (current ? { ...current, index } : current))}
+    />
+  ) : null;
+  const sidePanel = agentDetailPanel ?? documentPreviewPanel ?? activityPanel ?? threadPanel;
```
wanghaojie124, line 380: When `agentDetailPanelProps` is present, this precedence keeps the agent detail panel visible after `handlePreviewAttachment` stores a preview request. Clicking an attachment therefore appears to do nothing, and the stale preview opens only after the agent panel closes. Please either close the agent detail panel through its guarded `onClose` flow before accepting the request, or explicitly prevent/defer preview creation while that panel is active.

web/app/src/components/business/DocumentPreviewPanel/PowerPointPreview.tsx
```diff
@@ -0,0 +1,100 @@
...
+        return mainPreviewer.preview(copyPreviewBuffer(data));
+      })
+      .then(() => {
+        if (cancelled) {
+          return;
+        }
+        if (!mainPreviewer) {
+          throw new Error("PowerPoint preview is unavailable");
+        }
+        setSlideCount(mainPreviewer.slideCount);
```
wanghaojie124, line 51: The current `slide` state survives when `data` changes. If a user is on slide 10 and switches to a 3-slide deck, the controls display `10/3` and navigation keeps requesting out-of-range indexes, even though the preview initially renders slide 1. Reset `slide` to 1 (or clamp it to the new `slideCount`) when loading new data, and add a regression test for switching decks after navigation.

web/app/src/components/business/DocumentPreviewPanel/DocumentPreviewContent.tsx
```diff
@@ -0,0 +1,103 @@
...
+}: {
+  data: ArrayBuffer;
+  item: AttachmentPreviewItem;
+  objectURL: string;
+  scale: number;
+  t: (key: string) => string;
+}) {
+  const kind = documentPreviewKind(item, data);
+  const text = useMemo(
+    () => (kind === "markdown" || kind === "text" ? formatPreviewText(data, item.mediaType) : ""),
```
wanghaojie124, line 31: Text attachments are allowed up to 25 MiB, but this synchronously decodes the entire buffer on the main thread; JSON may then be parsed/stringified and source files are expanded again into Shiki HTML. A valid large generated log/source file can therefore freeze the chat tab or exhaust memory when previewed. Please enforce a substantially smaller text-preview limit (with truncation/download fallback), or move parsing/highlighting off the main thread.

internal/runtime/codex/appserver_manager.go
xxx7xxxx: Addressed in `d36ee858`. File-publishing capability provenance is now persisted alongside conversation mappings. A known file-capable thread can be resumed, while a loaded legacy thread or persisted mapping without provenance is replaced by a new thread started with `csgclaw_publish_file`. Regression coverage includes legacy `EnsureSession` followed by `EnsureEngineSession`, known-capable cold resume, and unknown persisted mappings.

web/app/src/pages/ConversationPage/components/ConversationPane/ConversationPane.tsx
xxx7xxxx: Addressed in `d36ee858`. Attachment preview now requests the agent-detail panel to close through its guarded `onClose(false)` path and aborts preview creation when closing is blocked, so no hidden stale preview is queued. A regression test covers the blocked-close case.

web/app/src/components/business/DocumentPreviewPanel/PowerPointPreview.tsx
xxx7xxxx: Addressed in `d36ee858`. PPTX preview now resets the active slide and slide count whenever the input data changes, then clamps the slide after the new deck is parsed. The regression test navigates to slide 10 and switches to a three-slide deck, which now starts at slide 1.

web/app/src/components/business/DocumentPreviewPanel/DocumentPreviewContent.tsx
xxx7xxxx: Addressed in `d36ee858`. Text, Markdown, and unknown-file text sniffing are now bounded to 256 KiB before decoding, JSON formatting, or syntax highlighting. Larger files show a truncation notice while retaining the original full-file download. Regression coverage verifies the preview is bounded and the complete file remains downloadable.

wanghaojie124:
LGTM

PR #546, opened 2026-09-02
feat(desktop): sync Windows taskbar icon with theme

youngbeom-shin:
## Summary

- update the Windows taskbar icon when the Profile color mode changes
- preserve the selected theme icon before the Electron window is created
- follow Windows color mode changes when the app uses the system theme


youngbeom-shin:
https://github.com/user-attachments/assets/fd6a5d50-1fc0-4cd2-ab32-c4be67ccee75

jialudev (commented):
Thanks for improving the Windows theme integration. I think we can keep the themed taskbar icon behavior, but the current AppUserModelID and relaunch changes introduce a few Windows integration risks that should be addressed before merging:

1. **Keep the AppUserModelID stable across themes.** `windowsTaskbarAppUserModelID()` changes the window identity between `.dark` and `.light`, while the installed Squirrel shortcut still uses `com.squirrel.csgclaw_desktop.CSGClaw`. Windows groups and pins taskbar items by AppUserModelID, so launching from an existing pinned shortcut can create a separate running button, and changing the theme can move the window into yet another taskbar identity. Different AppUserModelIDs are explicitly treated as different taskbar groups: https://learn.microsoft.com/en-us/windows/win32/shell/samples-appusermodelidwindowproperty

2. **Do not apply the Squirrel AppUserModelID to MSIX builds.** `main.ts` intentionally avoids overriding the application ID when `process.windowsStore` is true, but `applyWindowsTaskbarDetails()` now assigns a Squirrel ID to every Windows window, including MSIX/Store packages. MSIX applications must retain the AUMID generated from their package manifest; programmatically assigning a custom ID can break package/taskbar/Jump List identity and is incompatible with Microsoft's packaging guidance: https://learn.microsoft.com/en-us/windows/msix/desktop/desktop-to-uwp-prepare

3. **Do not use `process.execPath` as the Squirrel relaunch command.** In a Squirrel installation this points into a versioned directory such as `app-0.6.0/CSGClaw.exe`. A taskbar item pinned from this window can therefore launch an old version and eventually stop working after Squirrel removes that directory. Squirrel shortcuts deliberately use the stable `Update.exe --processStart CSGClaw.exe` entry point and update it across releases: https://github.com/Squirrel/Squirrel.Windows/blob/develop/docs/using/update-process.md

A safer implementation would separate presentation from identity:

- Cache the selected `NativeImage` before the `BrowserWindow` is created, as this PR already does.
- Always call `window.setIcon(icon)` when the theme actually changes.
- For non-Store/Squirrel builds, keep the existing stable `windowsAppUserModelID`, set only the themed `appIconPath`/`appIconIndex`, and leave the relaunch fields unset so the installed Squirrel shortcut remains the launch owner. If the skip-taskbar refresh is still needed, only perform it when the effective icon variant changes.
- For MSIX builds, do not call `setAppDetails()` with a custom `appId` or relaunch command; update the live window icon and let the package manifest own the pinned identity/icon.
- If Windows continues to prefer the installed icon for an already pinned shortcut, keep that pinned icon stable (or use an overlay) rather than changing the AppUserModelID. A theme-aware running-window icon is still useful without splitting the application identity.

It would also be valuable to validate these Windows-specific cases: a pinned Squirrel install before and after a theme change, the same pin after two application updates, and an installed MSIX build. The current shared theme test covers color selection but not the taskbar identity/relaunch behavior.

jialudev (approved):
LGTM

PR #553, opened 2026-09-04
feat(skill): add GitHub release publishing workflow skill

GatewayJ:
## Summary
- add a repository-scoped Skill for publishing stable and beta GitHub Releases
- target the latest `origin/main` commit and reuse `docs/releases/{version}.md` verbatim
- require exact user confirmation before `gh release create`; never run `git tag`, `git push`, or update `main`

## Validation
- `quick_validate.py .codex/skills/github-publish-release`
- read-only forward simulation stopped at the confirmation gate

## Notes
- remote tag, Release, and account prechecks are intentionally omitted
- no tag or Release was created while implementing this change

### DIFF
--- a/.codex/skills/github-publish-release/SKILL.md
+++ b/.codex/skills/github-publish-release/SKILL.md
@@ -12,10 +12,11 @@ Publish a GitHub Release for `OpenCSGs/csgclaw` through `gh release create`. The
 1. Require the user to provide both `version` and `tag`. Do not infer either value. Require them to be identical.
 2. Accept only a stable tag such as `v0.6.0` or a beta tag such as `v0.6.0-beta.1`. For alpha or any other format, stop without publishing.
 3. Run `git fetch origin main`, then resolve the full target commit with `git rev-parse origin/main^{commit}`. Do not switch branches or modify the worktree.
-4. Set the release-note path to `docs/releases/{version}.md`. Read it from the target commit with `git show {target_sha}:{note_path}`. If it is missing or empty, stop and report the missing file. Do not generate, rewrite, or supplement the release note.
-5. Show the repository, version, tag, release type, full target SHA, release-note path, and complete release-note body. Ask the user to reply exactly `确认发布 {tag} @ {short_sha}`. Stop and wait; no remote mutation is allowed before that exact confirmation.
-6. After confirmation, fetch `origin/main` again. If its full SHA differs from the previewed SHA, invalidate the confirmation, show the new preview, and require a new exact confirmation.
-7. Read the release-note body from the confirmed target commit and publish it through standard input:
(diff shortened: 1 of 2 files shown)


.codex/skills/github-publish-release/SKILL.md
```diff
@@ -0,0 +1,46 @@
...
+6. After confirmation, fetch `origin/main` again. If its full SHA differs from the previewed SHA, invalidate the confirmation, show the new preview, and require a new exact confirmation.
+7. Read the release-note body from the confirmed target commit and publish it through standard input:
+
+   Stable release:
+
+   ```bash
+   git show "${target_sha}:docs/releases/${version}.md" |
+     gh release create "${tag}" \
+       --repo OpenCSGs/csgclaw \
+       --target "${target_sha}" \
```
xxx7xxxx, line 27: Reject existing tags before publishing the confirmed commit

If the requested tag already exists on a different commit but has no Release, `gh release create --target "${target_sha}"` succeeds against that existing tag: GitHub ignores `target_commitish` when the tag exists. This publishes the confirmed main commit's notes against different code and then reports the confirmed SHA as if it were the actual release target. Check the remote tag before publication and stop if it exists, consistent with this skill's create-only contract. Apply the same guard to both stable and beta releases.

.codex/skills/github-publish-release/SKILL.md
```diff
@@ -0,0 +1,46 @@
+---
+name: github-publish-release
```
xxx7xxxx, line 2: Suggestion: put this shared skill under `.agents/skills/github-publish-release/` instead of `.codex/skills/github-publish-release/`. Codex automatically discovers repository skills in `.agents/skills/`, so this gives the workflow an agent-independent location while preserving Codex discovery. The optional `agents/openai.yaml` can remain inside the skill folder as Codex-specific metadata. See the [official skill discovery documentation](https://learn.chatgpt.com/docs/build-skills#where-to-save-skills).

GatewayJ: All existing skills in this repository are Codex-specific and are stored under .codex/skills, so this PR keeps the current repository convention instead of moving only this skill.

.codex/skills/github-publish-release/SKILL.md
GatewayJ: Fixed in bd904cd1. The skill now checks both the Git tag ref and Release for the requested tag before preview, and repeats both checks immediately before publishing. It stops if either exists.

.codex/skills/github-publish-release/SKILL.md
xxx7xxxx: We need to move them in future

xxx7xxxx (approved):

PR #557, opened 2026-09-05
refactor(agentengine): unify runtime ownership and extensions

xxx7xxxx:
## Summary

- Unify Agent management, hosted conversations, and RuntimeExtensions under shared Engine lifecycle and persistence.
- Manage Feishu lark-cli configuration through isolated, retryable extensions while preserving OpenClaw/PicoClaw native channels.
- Preserve extensions across runtime replacement and fix skipped answers, cancellation, and cleanup readiness, with regression tests and bilingual architecture documentation.

PR #558, opened 2026-09-07
feat(codex): support MCP workspace file uploads

wanghaojie124:
- add a turn-scoped dynamic tool for uploading workspace files to configured MCP servers
- validate workspace paths, upload sizes, content types, same-origin targets, redirects, and cancellation
- guard unresolved runtime input files before media-type handling
- localize the AGENT-ERR-25 pending sensitive-check response in the Web UI
- extend runtime and frontend coverage for the new behavior


xxx7xxxx:
Will handle this issue after #557 is merged.

xxx7xxxx:
@wanghaojie124 Please fix the conflict

wanghaojie124:
Resolved the conflicts against the latest `origin/main` and pushed the merge commit to the `dev` branch. The full `go test ./...` suite passes.

internal/runtime/codex/appserver_manager.go
```diff
@@ -940,13 +978,31 @@ func (m *appServerManager) handleAppServerDynamicToolCall(runtimeID string, live
...
+	if params.Tool == appServerUploadFileToolName {
+		var args appServerUploadFileArgs
+		if err := json.Unmarshal(params.Arguments, &args); err != nil {
+			return appServerDynamicToolResponse(false, "invalid upload arguments"), nil
+		}
+		uploadCtx, ok := live.appServerTurnContext(params.ThreadID)
+		if !ok {
+			return appServerDynamicToolResponse(false, "file upload is unavailable outside an active turn"), nil
+		}
+		if err := uploadAppServerWorkspaceFile(uploadCtx, live.spec, args); err != nil {
```
xxx7xxxx, line 1003: Enforce read-only mode before executing uploads

This branch executes host filesystem reads and HTTP PUTs even when `live.spec.ExecutionMode` is `read_only`. `EnsureEngineSession` registers the upload tool for read-only agents too, bypassing that mode's filesystem deny policy and rejection of external mutations. A local reproduction through the dynamic request handler successfully uploaded workspace contents from a read-only session. Reject uploads in read-only mode and omit this tool from its registration.

internal/runtime/codex/appserver_manager.go
```diff
@@ -940,13 +978,31 @@ func (m *appServerManager) handleAppServerDynamicToolCall(runtimeID string, live
...
 	if live == nil || !live.appServerPublishesFilesForThread(params.ThreadID) {
 		return nil, fmt.Errorf("dynamic file tool references unknown or unsupported thread %q", params.ThreadID)
 	}
-	file, err := normalizeAppServerPublishFileArgs(params.Arguments)
+	if params.Tool == appServerUploadFileToolName {
+		var args appServerUploadFileArgs
+		if err := json.Unmarshal(params.Arguments, &args); err != nil {
+			return appServerDynamicToolResponse(false, "invalid upload arguments"), nil
+		}
+		uploadCtx, ok := live.appServerTurnContext(params.ThreadID)
```
xxx7xxxx: Bind upload contexts to the requesting turn

Server requests run asynchronously, so a delayed request from canceled turn A can execute after turn B starts on the same thread. This lookup ignores `params.TurnID`, retrieves B's uncanceled context, and performs A's upload. A deterministic reproduction using the asynchronous dispatch path confirms that the HTTP PUT succeeds in this interleaving. Validate both thread and turn IDs and atomically retrieve the matching active context.

internal/runtime/codex/appserver_manager.go
```diff
@@ -967,6 +1023,74 @@ func (m *appServerManager) handleAppServerDynamicToolCall(runtimeID string, live
...
+	target, err := base.Parse(uploadURI)
+	if err != nil || target.Scheme != base.Scheme || !strings.EqualFold(target.Host, base.Host) || target.User != nil {
+		return fmt.Errorf("upload URI must be same-origin with MCP server %q", strings.TrimSpace(args.Server))
+	}
+	root, err := os.OpenRoot(spec.WorkspaceDir)
+	if err != nil {
+		return fmt.Errorf("open Runtime workspace: %w", err)
+	}
+	defer root.Close()
+	file, err := root.Open(cleaned)
```
xxx7xxxx: Reject non-regular files without a blocking open

`root.Open` blocks when the workspace path refers to a FIFO without a writer, before the regular-file check or HTTP timeout can apply. Canceling the turn also cannot interrupt that open. A local FIFO reproduction remains blocked after cancellation and returns only after a writer connects. Use a nonblocking open with descriptor validation, or another race-safe approach that rejects special files without waiting for a writer.

wanghaojie124: Fixed in 3d274c76. On non-Windows platforms the workspace file is opened through os.Root.OpenFile with O_NONBLOCK, then validated from the opened descriptor as a regular file. Added a FIFO regression test that verifies rejection completes without a writer or blocking.

internal/runtime/codex/appserver_manager.go
wanghaojie124: Fixed in 3d274c76. The active upload context is now stored with both the thread ID and turn ID, and lookup validates the pair atomically under liveSession.mu. Cleanup is also scoped to the matching turn so a stale turn cannot clear or reuse a newer context. Added a mismatched-turn regression test.

internal/runtime/codex/appserver_manager.go
wanghaojie124: Fixed in 3d274c76. Read-only sessions now omit csgclaw_upload_file from dynamic tool registration, and the request handler also rejects upload calls defensively. Added regression coverage for both behaviors.

internal/runtime/codex/appserver_manager.go
```diff
@@ -371,6 +384,7 @@ func (m *appServerManager) Prompt(ctx context.Context, handle SessionHandle, req
 		return PromptResponse{}, err
 	}
 	waiter.setTurnID(appServerTurnIDFromResult(raw))
+	live.setAppServerTurnContext(sessionID, waiter.currentTurnID(), turnCtx)
```
xxx7xxxx: Handle upload requests before the turn/start response arrives

The upload context is now registered only after the `turn/start` RPC returns, but `turn/started` and dynamic tool requests can be processed while that response is still pending. A valid upload in this window is rejected as `file upload is unavailable outside an active turn`, even though the waiter has already identified the active turn. A mock app-server process reproduces the failure on this head, while the same test passes on the previous head; it also occurs with the start response emitted immediately after the tool request. Bind the context when `turn/started` establishes identity, or defer the matching tool request until start initialization completes, while retaining stale-turn rejection.

wanghaojie124: Fixed in ea98fe90. The upload context and waiter are now registered before the turn/start request. Once turn/started assigns the waiter turn ID, an early matching tool request can retrieve the context immediately, while requests from stale or unidentified turns remain rejected. Cleanup is keyed by waiter identity, and regression coverage verifies availability immediately after turn/started.

xxx7xxxx (approved):

PR #562, opened 2026-09-08
feat(agentengine): batch agent skill summaries

xxx7xxxx:
## Summary

- Load Agent Profile skill names and descriptions in one Engine-backed request, keeping full file reads on demand.
- Bound metadata reads and isolate invalid files, with cancellation, stable ordering, and large-list regression coverage.

### DIFF
--- a/internal/api/agent_skills.go
+++ b/internal/api/agent_skills.go
@@ -23,7 +23,7 @@ func (h *Handler) handleAgentSkillSummaries(w http.ResponseWriter, r *http.Reque
 	}
 	item, err := h.agentEngine.Agents().Get(r.Context(), pathValue(r, "id"), agentengine.AgentGetOptions{IncludeSkillSummaries: true})
 	if err != nil {
-		writeAgentSkillsMutationError(w, err)
+		writeAgentGetError(w, err)
 		return
 	}
 	items := item.Status.SkillSummaries
--- a/internal/api/agent_skills.go
+++ b/internal/api/agent_skills.go
@@ -16,6 +16,23 @@ type batchAddAgentSkillsRequest struct {
 	Names []string `json:"names"`
 }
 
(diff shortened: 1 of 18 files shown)


xxx7xxxx:
As @GatewayJ suggested, we need optimize the batch fetching skills' performance.

internal/api/agent_skills.go
```diff
@@ -16,6 +16,23 @@ type batchAddAgentSkillsRequest struct {
...
 }
 
+func (h *Handler) handleAgentSkillSummaries(w http.ResponseWriter, r *http.Request) {
+	if h.agentEngine == nil {
+		http.Error(w, "agent engine is not configured", http.StatusServiceUnavailable)
+		return
+	}
+	item, err := h.agentEngine.Agents().Get(r.Context(), pathValue(r, "id"), agentengine.AgentGetOptions{IncludeSkillSummaries: true})
+	if err != nil {
+		writeAgentSkillsMutationError(w, err)
```
wanghaojie124: Could this read path use an error mapper that preserves server-side failures? `WorkspaceService.SkillSummaries` converts root/open/read failures into `Agent skill metadata is unavailable`, but `writeAgentSkillsMutationError` maps every unrecognized error to 400. A missing or unreadable runtime skills root is a server/storage failure rather than a bad client request, so this endpoint will return a misleading 400 and clients or monitoring may treat it as non-retryable. Mapping that failure to 500/503 (while retaining 404 for a missing agent) would keep the API semantics accurate.

xxx7xxxx: Fixed in 8f8b908b by reusing the Agent read-error mapper. Storage failures now return 500 with the sanitized message, while a missing agent remains 404. An absent skills directory still returns 200 with an empty list, since it represents no installed skills. Added HTTP regression coverage for all three cases and runtime-path redaction.

wanghaojie124:
LGTM

PR #563, opened 2026-09-08
fix(codex): isolate managed agents from host model providers

xxx7xxxx:
## Summary

- Remove unused host model providers when CSGClaw manages the Agent's model, preventing strict-config startup failures in read-only mode.
- Clean both new and existing Agent configurations without changing the host Codex file, with regression tests and real Codex startup verification.


wanghaojie124:
LGTM

PR #564, opened 2026-09-08
fix(server): avoid idle event stream shutdown delays

xxx7xxxx:
## Summary

- [#407](https://github.com/OpenCSGs/csgclaw/pull/407) (`ce7ff551`) correctly made shutdown await HTTP cleanup, but persistent event streams kept it waiting for the full five-second timeout whenever the Web UI was open.
- Close passive IM, Feishu, and participant event streams after desktop sandbox cleanup, preserving shutdown of both HTTP listeners, the five-second grace period for ordinary requests, and runtime cleanup.
- Allow a second Ctrl+C to terminate stalled cleanup in both foreground and daemon mode.
- Validated end to end with open event streams, in-flight request completion, desktop shutdown ordering, and child-process cleanup.

### DIFF
--- a/internal/server/http_test.go
+++ b/internal/server/http_test.go
@@ -2,6 +2,8 @@ package server
 
 import (
 	"context"
+	"errors"
+	"io"
 	"net"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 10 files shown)


xxx7xxxx:
@jialudev You may care this change since #407 was your contribution.

xxx7xxxx:
@wanghaojie124, please help review this pr

internal/server/http_test.go
```diff
@@ -11,9 +13,225 @@ import (
...
+					case <-time.After(6 * time.Second):
+						t.Error("server did not stop during cleanup")
+					}
+				}
+			})
+
+			client := &http.Client{Timeout: 10 * time.Second}
+			streamsDone := make(chan error, len(addresses))
+			for _, address := range addresses {
+				req, err := http.NewRequest(http.MethodGet, "http://"+address+"/api/v1/events", nil)
```
wanghaojie124, line 90: This regression test only opens `/api/v1/events`, so the new shutdown branches in the Feishu and participant event-stream handlers are not exercised. Those handlers have their own subscription cleanup, and the participant path also requeues buffered events on exit. Please add focused coverage that closes the shutdown channel for those two handlers and verifies that they return and preserve the expected cleanup/requeue behavior.

xxx7xxxx: Addressed in eb714183. Added focused tests for both handlers that close the shutdown channel and verify clean termination and subscription cleanup, including the participant control subscription. A separate buffered-event test verifies that undelivered messages are requeued and replayed exactly once after reconnect.

The tests also catch regressions when the shutdown or cleanup steps are removed. Verified locally, including race checks.

wanghaojie124:
LGTM

PR #565, opened 2026-09-09
feat(web): add sandboxed HTML attachment previews

xxx7xxxx:
## Summary

- Render HTML attachments with inline styles and a highlighted source view.
- Keep in-page links inside the preview while blocking scripts and external resources.

## Screenshots

image

image


jialudev (approved):
LGTM

PR #566, opened 2026-09-09
fix(web): default agent templates to codex

wanghaojie124:


web/app/src/pages/WorkspacePage/components/WorkspaceModals/AgentProfileModal.tsx
```diff
@@ -149,13 +149,7 @@ export function AgentProfileModal({
...
   const selectedProviderModels = selectedProvider?.models ?? [];
   const selectedModelValue = agentDraft.model_id || "";
-  const workerTemplates = useMemo(
-    () =>
-      workerSelectableTemplates(hubTemplates).filter(
-        (item) => normalizeRuntimeKind(item.runtime_kind) !== "picoclaw_sandbox",
-      ),
-    [hubTemplates],
-  );
+  const workerTemplates = useMemo(() => createAgentSelectableTemplates(hubTemplates), [hubTemplates]);
```
xxx7xxxx, line 152: Preserve explicitly selected Hub templates when filtering these options. The built-in OpenClaw worker remains visible in Hub, whose Create agent action passes it directly into this modal. This filter removes its option while leaving from_template populated, so the template selector is blank even though Create still submits OpenClaw. Switching to Custom and back also silently replaces that explicit selection with Codex. Keep the selected template available here, or consistently remove its Hub creation route.

wanghaojie124: Fixed in 1af6d100. The built-in OpenClaw worker no longer exposes the Create action in Hub, and openCreateAgentModal now defensively falls back to the built-in Codex template if that OpenClaw template is passed explicitly. Remote OpenClaw templates remain available. Added regression coverage for the Hub action and controller fallback.

xxx7xxxx (approved):

PR #567, opened 2026-09-09
feat: implement room-first manager collaboration

jialudev:
- Add on-demand rooms where Manager owns room conversations, plans work, dispatches tasks, and coordinates member contributions.
- Flatten task storage around parent IDs, move task history to an append-only event ledger, and normalize task metadata across room messages.
- Consolidate room task operations under the task CLI with automatic assignment resolution and validation.
- Refine collaboration-room defaults, speaking-mode behavior, task-plan presentation, and compact room-mode labels while treating rooms without a type as free discussion.


jialudev:
https://github.com/user-attachments/assets/f5f31938-84fa-4657-a101-ef609575b194


image

image

internal/roomtask/service.go
```diff
@@ -0,0 +1,575 @@
...
+			return fmt.Errorf("stop the active execution before reassignment")
+		}
+		if t.Status == taskcore.StatusCompleted || t.Status == taskcore.StatusReview {
+			return fmt.Errorf("review the submitted result before dispatching again")
+		}
+		if !dependenciesReady(a, t) {
+			return fmt.Errorf("waiting for predecessor acceptance by the manager")
+		}
+		for _, other := range tasks {
+			if other.ID != id && other.AssignedTo == assignee && (running(other) || other.RecoveryRequired) {
```
xxx7xxxx, line 369: Release or protect room task capacity when deleting a room. Deleting room A through DELETE /api/v1/rooms/{id} succeeds while its child is dispatched or in progress. That persisted child remains in this global capacity check, so dispatching the same worker in room B is permanently blocked. Updates/stops for A are rejected because its roster no longer exists, and Recover skips missing rooms, so restarting does not repair it. Reject deletion until execution is stopped, or close the room tasks and release their capacity as part of deletion.

cli/task/task.go
```diff
@@ -102,23 +121,49 @@ func (c cmd) runCreate(ctx context.Context, run *command.Context, args []string,
 }
 
 func (c cmd) runClaim(ctx context.Context, run *command.Context, args []string, globals command.GlobalOptions) error {
-	fs := run.NewFlagSet("task claim", run.Program+" task claim --task <id> --participant-id <participant>", "Claim an agent task.")
+	fs := run.NewFlagSet("task claim", run.Program+" task claim --task <id> --actor-id <participant> [--attempt <n>]", "Claim a task after resolving its assignment.")
 	taskID := fs.String("task", "", "task id")
-	participantID := fs.String("participant-id", "", "worker participant id")
+	actorID := fs.String("actor-id", "", "worker participant id")
```
xxx7xxxx, line 126: Update generated claim instructions together with this flag rename. internal/agenttask/service.go still emits task claim --task ... --participant-id ... for every newly created direct Agent task, and the current Codex/OpenClaw worker templates use the same command. This parser now accepts only --actor-id, so following those fresh instructions fails immediately with flag provided but not defined: -participant-id. Update the notification generator and worker templates to use the new flag.

internal/api/room_runtime_scope.go
```diff
@@ -0,0 +1,165 @@
...
+							selectedRoom = id
+						}
+					}
+				}
+			}
+		}
+		if selectedRoom == "" && len(rooms) == 1 {
+			selectedRoom = rooms[0]
+		}
+		if selectedRoom == "" {
```
xxx7xxxx: Exempt free/direct workflows before rejecting an ambiguous active-room set. ActiveRooms contains every active room, but this branch runs before checking IsOnDemand. When the same agent is active in two free rooms, even GET /api/v1/channels/csgclaw/rooms returns 409; the identical request succeeds with one free room. Room-specific message/history reads also cannot select a room here. Scope validation needs to distinguish the originating execution and apply the on-demand restriction only to that workflow.

internal/api/room_task_control.go
```diff
@@ -0,0 +1,106 @@
...
+		if t.ParentID != id || t.Status != taskcore.StatusInProgress {
+			continue
+		}
+		if !readable || !controllable {
+			problems = append(problems, "execution controller unavailable for "+t.ID)
+			continue
+		}
+		found := false
+		for _, l := range reader.ActiveWork(room) {
+			if h.roomTaskIDForSource(room, l.RequestID) != t.ID || h.roomTaskAttemptForSource(room, l.RequestID) != t.Attempt {
```
xxx7xxxx: Preserve task identity for user-input continuation leases. interaction.Observe resumes detached request_user_input answers with a synthetic structured-user-input-<id> MessageID. startWork stores that ID as the lease RequestID, but it exists in neither the task ledger nor IM messages, so both lookups here miss the active task. Stopping the parent then returns 409 awaiting execution confirmation without calling RequestStop, leaving the worker running. Retain a resolvable task source or carry the trusted task ID and attempt in the lease.

internal/taskcore/store.go
```diff
@@ -324,16 +439,21 @@ func buildTaskIndex(root string) ([]IndexEntry, error) {
 		if !entry.IsDir() {
 			continue
 		}
-		rootPath := filepath.Join(root, entry.Name(), rootFileName)
-		if _, err := os.Stat(rootPath); errors.Is(err, os.ErrNotExist) {
+		tasksPath := filepath.Join(root, entry.Name(), tasksFileName)
```
wanghaojie124, line 456: 这里改为只识别 `tasks.json`，但现有版本持久化的是 `index.json` 以及每个任务目录下的 `root.json`/`children.json`/`approvals.json`/`presence.json`。升级后 `buildTaskIndex` 会静默跳过全部历史任务；同时 `sequence.json` 不存在时 allocator 也会基于这个空索引从 0 开始，下一次创建可能复用 `task-1` 并把新格式写进旧目录。请在切换格式前迁移旧记录（并从旧 index/counter 保留序号），或至少兼容读取旧格式，并补一个从旧版 fixture 升级加载的回归测试。

internal/roomtask/service.go
```diff
@@ -0,0 +1,575 @@
...
+}
+func (s *Service) MembersChanged(room string) error {
+	for _, root := range s.List(room) {
+		if root.ParentID != "" || Terminal(root.Status) {
+			continue
+		}
+		if err := s.change(room, root.ID, func(r Roster, _ []taskcore.Task, a *taskcore.Snapshot, _ func() (string, error)) error {
+			for i := range a.Children {
+				t := &a.Children[i]
+				if !Terminal(t.Status) && t.Status != taskcore.StatusBlocked && !worker(r, t.AssignedTo) {
```
wanghaojie124: 这个条件也会命中 `review` 状态：Worker 已提交结果后如果被移出房间，任务会被改成 `blocked`。随后 `Review` 只接受 `review`（或带 result 的 `failed`），因此 Manager 无法再接受已保存的提交，只能重新派发。建议保留待审核状态/结果，仅对尚未提交的 pending、assigned、in_progress 任务做成员变更处理，并补覆盖该时序的测试。

GatewayJ:
房间删除接口直接删除 IM room，没有检查或停止未终结任务。PR 已经禁止删除正在执行任务的单个成员，但整个房间删除绕过了这一约束。

internal/roomtask/service.go
jialudev: Fixed in d036a1b1. Room deletion now acquires a room-task deletion guard and returns 409 while any task is unfinished or still requires recovery. New task mutations are blocked while deletion is in progress, with service and API regression coverage. The Web UI also keeps the confirmation dialog open and shows the localized conflict there.

cli/task/task.go
jialudev: Fixed in d036a1b1. The generated direct Agent task notification and both Codex/OpenClaw worker instruction templates now use task claim --actor-id, matching the unified task parser. Added CLI, notification, and embedded-template regression coverage.

internal/api/room_runtime_scope.go
jialudev: Fixed in d036a1b1. Runtime scope now resolves explicit room message/member requests first and applies ambiguity enforcement only when on-demand work is involved. Direct and free-room workflows no longer inherit on-demand restrictions merely because the Agent has other active leases. Added regression coverage for multiple free rooms and explicit free-room requests alongside on-demand work.

internal/api/room_task_control.go
jialudev: Fixed in d036a1b1. Work leases now carry the trusted task ID and attempt from TurnContext through the participant-work API and registry. Stop/recover logic uses that structured identity first and retains the source-message lookup as a fallback, so synthetic user-input continuation message IDs remain stoppable. Added lease, stop-event, and parent-stop regression coverage.

internal/taskcore/store.go
jialudev: Addressed the collision/corruption risk in d036a1b1. Historical task records are intentionally not migrated in this format change, but the new allocator now seeds its sequence from both the legacy index.json counter and existing task-N directories. New-format records therefore cannot reuse or write into a legacy task directory. Added a legacy-layout regression fixture that verifies the next ID and confirms the old directory remains untouched.

internal/roomtask/service.go
jialudev: Fixed in d036a1b1. Member removal now blocks only pending, assigned, or in-progress tasks. A submitted task in review keeps its result and remains reviewable after the worker leaves the room. Added regression coverage for that exact sequence through successful review.

jialudev:
@GatewayJ 这个问题与第一条 room task capacity Review 属于同一根因，已在 d036a1b1 一并修复：整房删除现在会先经过 room-task deletion guard；存在未终结或仍需恢复的任务时返回 409，并在删除进行期间阻止新的任务变更，因此不能再绕过单成员删除已有的生命周期约束。前端遇到该冲突时会保留删除弹窗，并显示本地化提示。相关 Service、API 和 UI 回归测试已补充。

GatewayJ (approved):

internal/taskcore/id_allocator.go
```diff
@@ -102,71 +107,67 @@ func (a *TaskIDAllocator) Bump(id string) error {
...
-	return nil
+	return writeTaskSequence(filepath.Join(a.root, sequenceFileName), taskSequenceState{LastTask: a.next})
 }
 
 func formatTaskIdentifier(value int64) string {
 	return fmt.Sprintf("task-%d", value)
 }
+
+// persistedTaskSequence preserves global task ID monotonicity without loading
+// legacy task records. This branch intentionally uses only tasks.json for task
```
wanghaojie124: 保留旧 counter 解决了 ID 复用和覆盖旧目录的问题，但没有解决升级兼容性：`Store.Load/buildTaskIndex` 仍只读取 `tasks.json`，所以当前版本已经持久化的 agent/team 任务（`root.json`、`children.json`、`approvals.json`、`presence.json`）升级后会从任务列表和 API 中静默消失，引用这些 task ID 的调度记录也无法再解析。除非项目明确接受丢弃现有用户任务状态，否则这里仍需要一次性迁移或兼容读取旧记录，并用旧版完整 snapshot fixture 验证升级后任务、事件、审批和 presence 均可访问。

internal/api/room_runtime_scope.go
xxx7xxxx: The all-free case is fixed, but a free/direct workflow still inherits restrictions when this Agent also has on-demand work. With active leases for one free room and one on-demand room, csgclaw-cli room list from the free-room workflow returns 409; the same actual CLI request succeeds when both rooms are free. Non-room task lookups and global management commands cannot select their originating room either. The request only identifies the Agent, so the current turn needs to be identified before applying the on-demand policy instead of classifying it using all of that Agent's active rooms.

internal/api/room_task_control.go
```diff
@@ -0,0 +1,120 @@
...
+		Assessment string `json:"assessment"`
+	}
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	room, id := pathValue(r, "id"), pathValue(r, "task_id")
+	if leases, ok := h.participantWork.(roomWorkReader); ok {
+		for _, l := range leases.ActiveWork(room) {
+			if h.roomTaskIdentityForWork(room, l) == id {
```
xxx7xxxx: Exclude the Manager coordination lease when checking whether the old Worker execution is still active. Recovery feedback carries the child task ID, and the Manager turn processing that feedback therefore has the same TaskID in its lease. With no Worker running, the Manager's own task recover request is rejected here with 409 because its current coordination turn matches the child. Filtering active leases by the task's assignee (or an explicit execution role) would keep real Worker executions protected without preventing the Manager from recording recovery.

jialudev: Fixed in 57db6b58. Recovery now resolves the task first and rejects only an active lease that matches the task ID, requested attempt, and actual task assignee. A Manager coordination lease carrying the same child task ID no longer causes a false 409, while the matching Worker execution is still protected. Added regression coverage for both cases.

internal/taskcore/id_allocator.go
jialudev: Fixed in 57db6b58. Store startup now performs an idempotent migration of legacy multi-file Agent/Team snapshots and the intermediate aggregate format into schema-versioned flat tasks.json plus events.jsonl, preserving root/child tasks, events, approvals, and presence. Legacy files are retained for rollback, and existing scheduled-run task references remain resolvable so they do not trigger duplicate work. The allocator also reconciles the old index counter, existing task directories, and every numeric task ID in migrated/current flat records, then immediately persists the maximum into a versioned canonical sequence.json. Added full legacy snapshot, scheduled-task, idempotency, conflict, and sequence regression coverage.

internal/api/room_task_control.go
```diff
@@ -0,0 +1,128 @@
...
+		if t.ParentID != id || t.Status != taskcore.StatusInProgress {
+			continue
+		}
+		if !readable || !controllable {
+			problems = append(problems, "execution controller unavailable for "+t.ID)
+			continue
+		}
+		found := false
+		for _, l := range reader.ActiveWork(room) {
+			if h.roomTaskIdentityForWork(room, l) != t.ID || h.roomTaskAttemptForWork(room, l) != t.Attempt {
```
xxx7xxxx: Apply the assignee check to Stop as well as Recover. A Worker is instructed to end its turn after asking a task-scoped question, while the Manager turn handling that question carries the same child task ID and attempt. This loop therefore treats the Manager coordination lease as the Worker execution. Reproducing that normal message route and stopping the parent returns 200, cancels the Manager lease, then rejects its completion with stale stop confirmation because the Manager is not the assignee; the child stays in_progress. Match the lease participant to the child assignee before requesting cancellation or counting an execution as found.

The previously reported [mixed-room scope restriction](https://github.com/OpenCSGs/csgclaw/pull/567#discussion_r3976036529) also remains unaddressed and still reproduces at this head. Please address both issues before merging.

wanghaojie124:
LGTM

internal/api/room_task_control.go
jialudev: Fixed in b73922ad. Stop and Recover now share the same execution-match predicate: task ID, attempt, and the task’s actual assignee must all match the active work lease before it is considered or stopped. A Manager coordination lease carrying the same child task ID/attempt is therefore ignored, while a real Worker execution remains protected. Added regression coverage for both the Manager-only lease path and the matching-Worker lease path.

internal/api/room_runtime_scope.go
jialudev: Fixed in b73922ad. I removed the global ActiveRooms-based request middleware because X-CSGClaw-Caller-Agent identifies the Agent, not the originating turn or room, so classifying a request from all of that Agent’s active leases was inherently ambiguous. Free/direct and global requests are no longer restricted by an unrelated on-demand-room lease. On-demand coordination is selected by the private per-turn room context, while concrete room-task mutations resolve the globally unique task ID and enforce the task Manager/assignee, attempt, and state. Message writes still require the sender to match the runtime caller. Regression tests cover global room listing with a runtime caller and sender spoofing.

xxx7xxxx (approved):

PR #576, opened 2026-09-15
feat(mcp): add secure OpenCSG gateway file bridging

wanghaojie124:
## Summary

- Add managed OpenCSG MCP Gateway support with runtime-only authentication.
- Route managed MCP requests through a trusted local CSGClaw proxy and dynamically inject the current user’s OpenCSG access token.
- Add a generic MCP file bridge that exposes workspace file paths through standard MCP tool calls.
- Validate file paths, MIME types, and size limits before encoding file content outside the model context.
- Scope file bridge credentials to each agent and MCP server using HMAC tokens.
- Rematerialize MCP configuration when creating, restoring, or updating Codex sessions.
- Use the loopback address for host-based Codex runtimes to avoid unintended HTTP proxy routing.
- Prevent local workspace files and skills from being incorrectly requested as MCP resources.
- Preserve the existing `llm_wiki` resolution and authentication flow.

## Security

- OpenCSG access tokens are not stored in community templates or exposed to the model.
- Template-provided MCP URLs are not trusted for managed Gateway connections.
- Base64 file content is transferred only between CSGClaw and the MCP service.
- File access is restricted to the corresponding agent runtime workspace.
- Internal file bridge credentials cannot be reused across agents or MCP servers.
- Upstream redirects are disabled to prevent credential forwarding to an unexpected host.

## Related Changes

- `file-parser`: adds the `parse_file_content` MCP tool for Base64-encoded document input.
- `resume-scorer`: routes document parsing through the managed Gateway and declares its file argument mapping with `file_bindings`.

## Validation

- Ran `go test ./...`.
- Added coverage for Gateway materialization, token forwarding, scoped file bridge authentication, workspace path validation, file encoding, and Codex runtime configuration.
- Verified the end-to-end resume parsing flow through a native MCP tool call.

## Impact

Existing `llm_wiki` MCP integrations are unaffected. Only MCP servers explicitly declaring:

`type: opencsg_mcp_gateway`

and:

`auth_type: csghub_access_token`

use the new Gateway flow.


internal/api/mcp_file_bridge.go
```diff
@@ -0,0 +1,242 @@
...
+		return
+	}
+	arguments := map[string]any{
+		mcpBindingString(binding, "filename_argument"): filename,
+		mcpBindingString(binding, "content_argument"):  base64.StdEncoding.EncodeToString(content),
+	}
+	if key := mcpBindingString(binding, "content_type_argument"); key != "" {
+		arguments[key] = contentType
+	}
+	result, err := callMCPFileBridgeTool(ctx, serverName, map[string]any{
```
xxx7xxxx: Disable redirects on the file-bridge upstream client. This path calls internal/mcp.CallTool directly, whose probeHTTPClient follows redirects, so it bypasses the no-redirect client used by handleOpenCSGMCPGatewayProxy. A 307/308 response to tools/call replays the Base64 document body to another origin even when Authorization is stripped. I reproduced this through a real MCP client and the bridge HTTP route: a second HTTP server received the workspace document after a 307. Use a client that rejects redirects for this authenticated file-transfer path.

internal/runtime/codex/runtime.go
```diff
@@ -165,6 +168,40 @@ type Dependencies struct {
...
+		marker := "/api/v1/opencsg-mcp-gateway/mcp"
+		if !strings.HasSuffix(strings.TrimRight(urlText, "/"), marker) {
+			continue
+		}
+		base := strings.TrimSuffix(strings.TrimRight(urlText, "/"), marker)
+		entry["url"] = base + "/api/v1/agents/" + url.PathEscape(agentID) + "/mcp-file-bridge/" + url.PathEscape(serverName)
+		headers, _ := entry["headers"].(map[string]any)
+		authorization, _ := headers["Authorization"].(string)
+		rootToken := strings.TrimSpace(strings.TrimPrefix(authorization, "Bearer "))
+		headers["Authorization"] = "Bearer " + opencsgmcp.FileBridgeToken(rootToken, agentID, serverName)
```
xxx7xxxx, line 205: Allow the scoped bridge credential through the desktop listener before switching Codex to it. Desktop startup sets cfg.Server.ListenAddr to the sandbox listener, so the generated bridge URL targets desktopSandboxSecurityHandler. That middleware accepts only the root server token and rejects this HMAC token before handleMCPFileBridge can validate it. I reproduced a 401 with WWW-Authenticate: Bearer realm="csgclaw-desktop" through server.Run using the generated credential. Consequently, MCP file tools cannot initialize in the Electron app. Route the bridge through an appropriate endpoint or add narrowly scoped token validation to the desktop middleware.

internal/api/mcp_file_bridge.go
wanghaojie124: Resolved in 621e1c8e. The file bridge now uses an MCP transport that rejects HTTP redirects for both tools/list and tools/call, so authenticated requests and Base64 file content cannot be replayed to a redirect target. A regression test exercises a real 307 tools/call response and verifies that the target server receives no request.

internal/runtime/codex/runtime.go
wanghaojie124: Resolved in 621e1c8e. The desktop security middleware now recognizes the scoped HMAC bridge credential only for POST /api/v1/agents/{agent}/mcp-file-bridge/{server}, while continuing to require the root token for other routes. Regression tests cover the matching route and reject reuse against another agent, server, method, or ordinary API endpoint.

xxx7xxxx (approved):

PR #577, opened 2026-09-15
feat: add dedicated image generation models

xxx7xxxx:
## Summary

- Add optional Agent image models with faithful prompts, persistent image delivery, targeted retries, and image copying in Codex web conversations.
- Discover image models from provider metadata, improve Images API compatibility and error reporting, and refresh CLIProxy's GPT model catalog.

## Screenshots

image

image

image


internal/agentengine/engine.go
```diff
@@ -168,6 +171,30 @@ func (c *conversations) Run(ctx context.Context, request TurnRequest, sink Event
...
+		if request.ImageGeneration.Model == nil && c.engine.agents != nil {
+			selected, err := c.engine.agents.Get(turn.ctx, c.agentID, AgentGetOptions{})
+			if err == nil {
+				task := *request.ImageGeneration
+				task.Model = selected.Spec.Model.ImageGeneration
+				request.ImageGeneration = &task
+			}
+		}
+		err := c.generateImage(turn.ctx, turn, sink, *request.ImageGeneration)
+		result := TurnResult{Status: TurnSucceeded}
```
wanghaojie124: Successful image-only turns leave `Dispatched` false. `generateImage` has already registered the generated snapshot in `FileStore`, but `complete` only puts dispatched turns into the bounded completed-turn cache whose eviction calls `DeleteTurn`. A provider-failure retry that later succeeds therefore retains up to 32 MiB forever, so repeated retries can exhaust local disk. Please either mark successful image-only turns dispatched (while keeping failed ones retryable) or explicitly delete the stored file after confirmed IM delivery.

xxx7xxxx: Fixed in 2e33699f. Successful image-only turns now set `Dispatched: true`, so they enter the bounded completed-turn cache and eviction deletes their temporary snapshots. Failed attempts remain retryable.

Added regression coverage for a provider failure followed by a successful retry, replay without regeneration, and snapshot cleanup on eviction. Validated end to end.

wanghaojie124:
LGTM

PR #581, opened 2026-09-16
feat: add room attachment discovery and download

xxx7xxxx:
## Summary

- Let room members find and download historical attachments while keeping agent workspaces isolated.
- Preserve inherited file inputs and guide Codex/OpenClaw agents through attachment retrieval.
- Validated end to end: manager/worker handoff, history lookup, and same-name file versions.


wanghaojie124:
LGTM

PR #587, opened 2026-09-18
fix(web): prevent template details from covering agent creation

youngbeom-shin:
## Summary
- Keep template detail visibility tied to the selected template ID so data refreshes cannot reopen it over agent creation.
- Add local and remote template regressions covering asynchronous creation, usable form input, and reopening details.


jialudev (approved):
LGTM

PR #590, opened 2026-09-20
feat(connectors): support global GitLab PAT connections

wanghaojie124:
- Store and manage GitLab Connector credentials globally while keeping per-agent App bindings.
- Support configurable GitLab MCP endpoints, request headers, legacy credential migration, and automatic session refresh.
- Update CLI, API, desktop, and web configuration flows—including the global Apps entry—along with tests and documentation.


internal/api/connectors.go
```diff
@@ -273,16 +271,28 @@ func (h *Handler) handleGitLabConnectorConfig(w http.ResponseWriter, r *http.Req
...
-	status, err := svc.Disconnect(r.Context(), connectors.ProviderGitLab)
+	var (
+		status connectors.Status
+		err    error
+	)
+	switch r.Method {
+	case http.MethodGet:
+		status, err = svc.GitLabStatusForAgent(agentID)
+	case http.MethodPost:
+		status, err = svc.DisconnectGitLabForAgent(agentID)
```
xxx7xxxx: Invalidate the Agent's live GitLab App connections when its Connector is disconnected or updated. This handler only deletes the persisted Connector state, while buildTransport has already copied the PAT into each MCP client's static headers. Reproduced through the App/Connector HTTP routes and a real MCP client: rotating the PAT still sent the old PAT, and after disconnect returned connected=false, tools/call continued succeeding with that same PAT. Close/detach dependent MCP sessions on disconnect and refresh them when the Connector changes, including changes saved through another App.

internal/apps/connector.go
```diff
@@ -442,6 +510,54 @@ func (s *Service) open(ctx context.Context, agentID string, c Config, credential
...
+		"custom_header_count", len(credentials.Headers),
+	}
+	if detail != nil {
+		attrs = append(attrs,
+			"error_code", detail.Code,
+			"upstream_status", detail.HTTPStatus,
+			"authentication_error", detail.Authentication,
+		)
+	} else {
+		message := redactMCPLogMessage(err.Error(), c, credentials)
```
xxx7xxxx: Avoid logging raw upstream errors without redacting resolved Connector credentials. For a saved GitLab App, credentials.Token is deliberately empty and the actual PAT is obtained inside buildTransport, so redactMCPLogMessage cannot remove it. A real MCP tools/list response containing a JSON-RPC error that echoes PRIVATE-TOKEN writes the complete managed PAT to the server log; I reproduced this with the saved-credential resolver path. Use safe structured errors or redact all credentials actually injected into the transport, including Connector and platform-login tokens.

xxx7xxxx:
Please fix the conflict since the main has changed a lot

internal/api/connectors.go
wanghaojie124: Resolved in `7a36da7f`. GitLab Connector updates now call `RefreshConnector`, which detaches and closes every dependent MCP session across agents before reconnecting installations that should remain connected. Disconnect calls `InvalidateConnector`, immediately removing the affected tools and closing the transports so they cannot keep using the old PAT. `TestRefreshConnectorRotatesAllAgentSessionsAndInvalidatesOnDisconnect` covers both token rotation and disconnect behavior.

internal/apps/connector.go
wanghaojie124: Resolved in `7a36da7f`. Connection logging no longer includes raw SDK/upstream errors when there is no safe structured `ConnectionError`; it records a generic diagnostic plus non-secret metadata instead. This covers credentials resolved below the persisted App credentials, including Connector PATs and platform-login tokens. `TestConnectionFailureLogDoesNotIncludeUntrustedUpstreamError` verifies that an upstream error echoing `PRIVATE-TOKEN` cannot leak the managed PAT.

internal/api/apps.go
```diff
@@ -250,6 +271,50 @@ func (h *Handler) handleAgentApp(w http.ResponseWriter, r *http.Request) {
...
+		BaseURL: config.GitLabBaseURL, AccessToken: credentials.Token,
+	}); err != nil {
+		return noop, err
+	}
+	if h.apps != nil {
+		if err := h.apps.RefreshConnector(ctx, connectors.ProviderGitLab); err != nil {
+			slog.Warn("refresh GitLab Apps after Connector update", "agent_id", agentID, "error", err)
+		}
+	}
+	credentials.Token = ""
```
xxx7xxxx, line 304: Clear the persisted App token when updating a resource to Connector authentication. Setting the incoming credentials.Token to an empty string does not remove the previous token: updateResource calls mergeCredentials, which keeps non-empty stored values, and unlike Create it never clears them for connector mode. Reproduced by editing a GitLab App that already has a bearer token and saving the new Connector settings with a different PAT: the global Connector contains the new PAT, but the resource still reports credentials_set.token=true and a real MCP tool call sends the old App token because buildTransport prefers it over ResolveConnectorHTTP. Strip App-owned token/AppID/AppSecret/Env fields in the resource update path after merging, as Create already does, so saved Connector Apps always use the managed PAT.

wanghaojie124: Resolved in `c3bfb7a5`. Connector-mode resource updates now clear any previously stored App-owned token, App ID, App Secret, and environment credentials after merging the update, matching the Create path. Added `TestUpdateResourceToConnectorClearsAppOwnedCredentials` to verify that `credentials_set` no longer reports the stale values and that a real MCP tool call uses the managed Connector PAT instead of the previous App token.

xxx7xxxx (approved):

PR #592, opened 2026-09-20
fix: tune task toolbar action buttons

youngbeom-shin:
## Summary
- show task toolbar tooltips below the action buttons to avoid covering adjacent controls
- make create task and create scheduled task buttons primary actions


youngbeom-shin:
before：

https://github.com/user-attachments/assets/16e76fe9-6224-4e04-bff3-7a778baa84df


after：

https://github.com/user-attachments/assets/2bd90f2e-efed-4851-8638-3ba79ce2a1d2

jialudev (approved):
LGTM

PR #593, opened 2026-09-20
fix(agent): speed up skill preservation during recreate

GatewayJ:
## Summary

- preserve custom Skills during recreate with recoverable same-filesystem rename transactions
- restore preserved Skills after Codex startup seeding and recover interrupted initial journals

## Validation

- `go test ./internal/agentengine/agents ./internal/api`
- `go test -race ./internal/agentengine/agents`
- `go vet ./...`


internal/agentengine/agents/controller.go
```diff
@@ -774,14 +850,29 @@ func (s *Controller) persistManagerAgent(ctx context.Context, manager Agent, run
 
 func (s *Controller) cleanupBootstrapManagerForRecreate(ctx context.Context, rt sandbox.Runtime, runtimeHome, runtimeKind string) (sandbox.Runtime, error) {
 	log.Printf("force recreating bootstrap manager box %q", ManagerName)
+	managerHome, err := s.agentHomeDir(ManagerUserID)
+	if err != nil {
+		return rt, err
+	}
+	sourceRuntimeKind := s.managerSkillPreservationSourceRuntimeKind(runtimeKind)
+	preservedSkills, err := s.prepareWorkspaceSkillsPreservation(ManagerUserID, sourceRuntimeKind, runtimeKind, RoleManager)
```
wanghaojie124: Skills preservation now runs before the bootstrap manager box is removed or the runtime is closed. Because preparation renames the current skills root and creates a new one, the still-running manager can write into that replacement root; `removeAll(managerHome)` then deletes those writes, causing changes made during this window to be lost. Please quiesce/remove the runtime before starting preservation, as the previous ordering did, or otherwise prevent workspace writes during the transaction.

GatewayJ: Good catch. This helper belonged to the legacy bootstrap-sandbox manager path and no longer had any callers after manager recreation moved to ensureCodexManager. I removed cleanupBootstrapManagerForRecreate and its private runtime-kind helper, eliminating this dormant write-loss window. The active manager recreation path already calls runtimeImpl.Stop before prepareWorkspaceSkillsPreservation, and deletes the old runtime only after preservation succeeds. Validated with go test ./... and go vet ./....

internal/agentengine/agents/configuration.go
```diff
@@ -1114,63 +1104,99 @@ func (s *Controller) recreate(ctx context.Context, id string, imageFor func(cont
...
 	}
-	if restoreSkills != nil {
-		if err := restoreSkills(); err != nil {
+	if skillsPreservation != nil {
+		if err := skillsPreservation.Restore(); err != nil {
 			return Agent{}, fmt.Errorf("restore runtime skills: %w", err)
 		}
 	}
-
+	handle, err = runtimeImpl.New(ctx, createSpec)
```
xxx7xxxx, line 1187: Restoring the preserved skills before Runtime.New allows Codex's startup seeding to delete them again. ensureSession calls seedCodexHomeSkillsForExecutionMode and seedCodexHomeWorkspaceSkills, which remove same-named target directories before copying the host/workspace versions. An agent-local edit to an imported custom skill is therefore silently lost on a successful recreate, and Commit then removes the transaction. I reproduced this through RecreateRecord with the real Codex runtime and mocked process startup: the agent's SKILL.md becomes the host original; the same test passes on the parent commit. Ensure custom skills survive runtime seeding before committing the transaction, including the equivalent manager path.

internal/agentengine/agents/workspace_skills_transaction.go
```diff
@@ -0,0 +1,898 @@
...
+	data, err := os.ReadFile(statePath)
+	if errors.Is(err, os.ErrNotExist) {
+		entries, readErr := os.ReadDir(tempRoot)
+		if readErr != nil {
+			return readErr
+		}
+		if len(entries) == 0 {
+			return os.Remove(tempRoot)
+		}
+		return fmt.Errorf("skills transaction %q has no state file", tempRoot)
```
xxx7xxxx, line 269: An interruption during the very first persistState leaves state.json.tmp but no state.json. At that point no skills have been moved, yet this branch rejects the nonempty transaction directory, and NewController propagates the error on every subsequent startup. I reproduced startup failing with a transaction directory containing only a partial state.json.tmp. Recognize and safely clean this pre-journal state when no preservation data exists, and cover interruption before the initial journal rename.

wanghaojie124:
LGTM

internal/agentengine/agents/workspace_skills_transaction.go
GatewayJ: Fixed in bc71dabf. Recovery now removes a pre-journal directory only when its sole entry is a regular state.json.tmp; unknown or preservation data still returns an error. Added recovery and rejection tests.

internal/agentengine/agents/configuration.go
GatewayJ: Fixed in bc71dabf. Restore now runs immediately after Runtime.New in both agent and manager recreate paths, so Codex host and workspace seeding completes before preserved edits are restored. Added real Codex runtime regression tests for both paths.

PR #595, opened 2026-09-21
perf(modelprovider): validate responses on first SSE event

wanghaojie124:
- Return as soon as the first valid Responses API stream event arrives instead of waiting for response completion.
- Preserve immediate upstream failure mapping and update streaming validation coverage.
- 
image


internal/modelprovider/openai.go
```diff
@@ -317,26 +316,23 @@ func decodeOpenAIResponsesProbeStream(r io.Reader) (openAIResponsesProbeResponse
...
-			}
-			completed = event.Response
-			return true, nil
 		case "response.failed", "error":
 			code := UpstreamErrorCode([]byte(data))
 			status := UpstreamStatusForErrorCode(code)
 			return false, &ResponsesAPIStatusError{Operation: "responses", Status: http.StatusText(status), StatusCode: status, Body: data}
 		}
-		return false, nil
+		if eventType == "" {
```
xxx7xxxx: Only accept a valid Responses API event before marking this probe successful. Currently any nonempty event name succeeds: an upstream stream containing `event: ping` with `data: {}`, followed by `response.failed`, returns nil on the ping and the Codex configuration validator caches that provider/model as valid. The parent revision skips the ping and reports the failure. Ignore unrelated SSE events and validate the recognized response event before returning success.

@wanghaojie124 Since this PR is already merged, please address this in the next PR if it has not already been fixed.

internal/modelprovider/openai.go
```diff
@@ -283,7 +283,7 @@ func CheckResponsesAPIWithClient(ctx context.Context, client *http.Client, baseU
 	var probe openAIResponsesProbeResponse
 	mediaType := strings.ToLower(strings.TrimSpace(strings.Split(resp.Header.Get("Content-Type"), ";")[0]))
 	if mediaType == "text/event-stream" {
-		probe, err = decodeOpenAIResponsesProbeStream(resp.Body)
+		return validateOpenAIResponsesProbeStream(resp.Body)
```
xxx7xxxx: Preserve the UpstreamRequestError wrapper when returning stream parsing or transport failures. This direct return bypasses the wrapper below, so UserFacingUpstreamError no longer recognizes malformed SSE or a timeout after headers. Through ValidateConfig and writeAgentOperationError, these now become raw HTTP 400 responses containing the provider URL/model and diagnostics instead of the previous structured 502 upstream_unavailable or 504 upstream_timeout responses. Wrap non-nil stream errors while retaining the immediate success return.

@wanghaojie124 Since this PR is already merged, please address this in the next PR if it has not already been fixed.

PR #596, opened 2026-09-21
fix(agent): Copy host skills when creating or rebuilding a Codex agent

GatewayJ:
Copy host skills when creating or rebuilding a Codex agent. Starting, restarting, or restoring an existing agent preserves its current host skill copies while continuing to synchronize template and built-in skills. Restoring the manager during service startup also preserves its current skill set.

Adding or removing skills from a running Codex agent automatically executes `Stop → Provision → Start` so app-server reloads the skill set. Stopped agents apply changes on their next start. `Provision` runs the configured credential preparation and `InitShell`.

Validation:

- Added tests for skill additions and removals, unchanged lists, stopped agents, restart failures, and worker/manager startup, restoration, and rebuilding.
- `make fmt` and `make build-web` passed.
- `env -u CSGHUB_API_BASE_URL go test ./...` passed. The inherited Hub URL override was removed from the test process to use the default URL expected by the tests.


internal/agentengine/agents/resource_controller.go
```diff
@@ -478,6 +480,14 @@ func (f *Controller) updateDesired(ctx context.Context, agentID, resourceVersion
 				return err
 			}
 		}
+		if skillsChanged && !replacesRuntime && !forceRecreate &&
```
wanghaojie124: A request that changes both skills and a restart-requiring runtime setting can restart the Codex runtime twice: `UpdateRecord` already performs an automatic restart for such settings, then this branch restarts it again after applying the skills. That runs provisioning/`InitShell` twice and causes two avoidable app-server interruptions. Consider deferring the settings restart until after skill mutation, or tracking whether `UpdateRecord` already restarted and folding both changes into one lifecycle cycle; a combined-update regression test would make this explicit.

jialudev (approved):
LGTM

